PR #1, opened 2020-10-16
Migrate PyTorch backend from server

CoderHam:
- Uses new Tritonbackend API
- Add deps and reorganize backend code
- Add c10 and mkl deps to pytorch build


src/libtorch.cc
```diff
@@ -0,0 +1,1122 @@
...
+      const std::vector<torch::Tensor>& output_tensors,
+      TRITONBACKEND_Request** requests, const uint32_t request_count,
+      std::vector<TRITONBACKEND_Response*>* responses);
+
+  ModelState* model_state_;
+
+  // The full path to the TorchScript model file.
+  std::string model_path_;
+
+  std::shared_ptr<torch::jit::script::Module> torch_model_;
```
deadeyegoodwin, line 262: I think the segfault is caused by torch_model_ being null.
Here I think you can use unique_ptr... there is no need to have multiple owners for torch_model_.

src/libtorch.cc
```diff
@@ -0,0 +1,1122 @@
...
+    ModelState* model_state, TRITONBACKEND_ModelInstance* triton_model_instance)
+    : BackendModelInstance(model_state, triton_model_instance),
+      model_state_(model_state), device_(torch::kCPU)
+{
+  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+    device_ = torch::Device(torch::kCUDA, DeviceId());
+  }
+
+  THROW_IF_BACKEND_INSTANCE_ERROR(model_state->LoadModel(
+      ArtifactFilename(), device_, &model_path_, torch_model_));
```
deadeyegoodwin, line 303: Need to pass torch_model_ by reference ("&torch_model_")

src/libtorch.cc
```diff
@@ -0,0 +1,1122 @@
...
+    : BackendModel(triton_model)
+{
+  // TODO Obtain backend configuration and validate
+}
+
+TRITONSERVER_Error*
+ModelState::LoadModel(
+    const std::string& artifact_name, const torch::Device device,
+    std::string* model_path,
+    std::shared_ptr<torch::jit::script::Module> torch_model)
```
deadeyegoodwin, line 124: std::unique_ptr<...>* torch_model.
As it is now torch_model goes out of scope at end of this function and so releases.

src/libtorch.cc
```diff
@@ -0,0 +1,1122 @@
...
+            "' for model instance '" + Name() + "'");
+  }
+
+  // Serialize the torch model to string
+  std::string model_data_str;
+  RETURN_IF_ERROR(ReadTextFile(*model_path, &model_data_str));
+
+  try {
+    std::istringstream model_stream(model_data_str);
+    torch_model = std::make_shared<torch::jit::script::Module>(
```
deadeyegoodwin, line 152: Don't create a new shared (unique) ptr, you juse want to torch_model->reset(...)

deadeyegoodwin (changes requested):

src/libtorch.cc
CoderHam: I used a shared pointer because torch::jit::load() returns a shared pointer i.e. `std::shared_ptr<torch::jit::script::Module>` and not `torch::jit::script::Module`

deadeyegoodwin (changes requested):

src/libtorch.cc
```diff
@@ -0,0 +1,1113 @@
...
+        ("failed to load model '" + Name() + "': " + ex.what()).c_str());
+  }
+
+  return nullptr;  // success
+}
+
+TRITONSERVER_Error*
+ModelState::AutoCompleteConfig()
+{
+  // If the model configuration already specifies max_batch_size,
```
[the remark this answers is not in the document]
CoderHam: No we cannot. They only thing we can do is autocomplete the default_model_filename. Should we simply return an error saying that AutoCompleteConfig is not supported for PyTorch models?

src/libtorch.cc
```diff
@@ -0,0 +1,1113 @@
...
+        triton_model, 1 /* config_version */, message));
+  }
+
+  return nullptr;  // success
+}
+
+ModelState::ModelState(TRITONBACKEND_Model* triton_model)
+    : BackendModel(triton_model)
+{
+  // TODO Obtain and validate backend configuration
```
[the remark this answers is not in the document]
CoderHam: You're right. Removed.

src/libtorch.cc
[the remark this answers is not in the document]
deadeyegoodwin: No, you just can't do anything... you could log a warning saying "model configuration auto-complete is not supported for model 'blah'".
 Using default_model_filename is not really auto-complete, and you are handling that correctly elsewhere in the file.

src/libtorch.cc
```diff
@@ -0,0 +1,1113 @@
...
+// TRITONBACKEND_Model.
+//
+class ModelState : public BackendModel {
+ public:
+  static TRITONSERVER_Error* Create(
+      TRITONBACKEND_Model* triton_model, ModelState** state);
+  virtual ~ModelState() = default;
+
+  // Load a TorchScript model using 'artifact_name' as the name for the
+  // TorchScript file/directory. If 'instance_group_kind' is not
```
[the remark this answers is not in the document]
CoderHam: Removed.

src/libtorch.cc
```diff
@@ -0,0 +1,1089 @@
...
+    if (max_batch_size != 0) {
+      batchn_shape[0] = total_batch_size;
+    }
+
+    // The input must be in contiguous CPU/GPU memory.
+    const int64_t batchn_byte_size = GetByteSize(input_datatype, batchn_shape);
+
+    std::initializer_list<BackendMemory::AllocationType> alloc_perference;
+    if (device_.is_cpu()) {
+      alloc_perference = {BackendMemory::AllocationType::CPU_PINNED_POOL,
```
deadeyegoodwin, line 791: If the model is expecting input in CPU memory then don't use pinned. That is only if we needed to first stage some CPU data before transferring it to GPU.

deadeyegoodwin (changes requested):

src/libtorch.cc
CoderHam: Fixed

deadeyegoodwin (approved):

PR #15, opened 2021-03-09
Disable Werror during torchvision vision header

CoderHam:
- extern initialzied while being declared in vision.h
PS: Needed by 21.03 container since it picks up the new Torchvision changes that cause this issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,7 +218,9 @@ else()
   )
 endif() # TRITON_PYTORCH_DOCKER_BUILD
 
-# Need to turn off -Werror due to Torchvision vision header extern initialization
+# Need to turn off -Werror due to Torchvision vision.h extern initialization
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -218,11 +218,12 @@ else()
   )
 endif() # TRITON_PYTORCH_DOCKER_BUILD
 
+# Need to turn off -Werror due to Torchvision vision header extern initialization
```
deadeyegoodwin, line 221: Add "Unfortunately gcc does not provide a specific flag to ignore the specific warning: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=45977"

deadeyegoodwin (changes requested):
Remove WIP and I will merge.

deadeyegoodwin (approved):

PR #16, opened 2021-03-11
Update build to correctly handle RUNPATH and dependencies

deadeyegoodwin:


CoderHam:
These changes look good to me. Still waiting on the CI but I wanted to confirm that the opencv libs are loaded correctly when needed.

CoderHam (approved):

PR #21, opened 2021-03-30
Improve error messages to be more clear

CoderHam:


src/libtorch.cc
```diff
@@ -884,10 +884,10 @@ ModelInstanceState::ReadOutputTensors(
...
           TRITONSERVER_ErrorNew(
               TRITONSERVER_ERROR_INVALID_ARG,
-              (std::string("unexpected datatype TYPE_") +
-               TRITONSERVER_DataTypeString(output_dtype) +
-               " for inference output '" + name + "', expecting TYPE_" +
-               TRITONSERVER_DataTypeString(config_datatype))
+              (std::string("configuration expects datatype TYPE_") +
+               TRITONSERVER_DataTypeString(config_datatype) + " for output '" +
+               name + "', model provides TYPE_" +
+               TRITONSERVER_DataTypeString(output_dtype))
```
GuanLuo, line 890: You don't need to make the change here, this is inside model execution so the config is in sync with the model.

Actually, is this check even necessary? There is no way to check the model output type unless you actually run an inference on the model?

deadeyegoodwin: I believe pytorch can produce different output types for any given inference request. You could argue that we shouldn't check anyway and just return what it produces but that is inconsistent with how we handle other frameworks. If that is the case I think the change is ok.

GuanLuo: I see. I agree that having this check will be consistent with other frameworks.

GuanLuo (approved):

PR #22, opened 2021-05-05
Fix timing for GPU execution

tanmayv25:
### Before 
`Avg request latency: 138887 usec (overhead 42 usec + queue 80 usec + compute input 13897 usec + compute infer 10288 usec + compute output 114580 usec)`

### After
`Avg request latency: 128321 usec (overhead 36 usec + queue 75 usec + compute input 20 usec + compute infer 128107 usec + compute output 83 usec)`

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -678,7 +678,7 @@ ModelInstanceState::ProcessRequests(
     // Attaching callback on the stream ensures correct timestamp
     // is captured.
     cudaLaunchHostFunc(
-        c10::cuda::getCurrentCUDAStream().stream(), TimestampCaptureCallback,
+        torch::cuda::getCurrentCUDAStream(), TimestampCaptureCallback,
         reinterpret_cast<void*>(&compute_end_ns));
   } else {
     SET_TIMESTAMP(compute_end_ns);
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -55,6 +55,18 @@
 
 namespace triton { namespace backend { namespace pytorch {
 
+namespace {
+
+#if defined(TRITON_ENABLE_GPU) && defined(TRITON_ENABLE_STATS)
+void CUDART_CB
+TimestampCaptureCallback(void* data)
+{
+  SET_TIMESTAMP(*(reinterpret_cast<uint64_t*>(data)));
+}
+#endif
+
+}  // namespace
+
 //
 // ModelState
 //
@@ -660,7 +672,14 @@ ModelInstanceState::ProcessRequests(
   Execute(&responses, request_count, &input_tensors, &output_tensors);
 
   uint64_t compute_end_ns = 0;
-  SET_TIMESTAMP(compute_end_ns);
+#if defined(TRITON_ENABLE_GPU) && defined(TRITON_ENABLE_STATS)
+  // For GPU, the Execute runs the inference asynchronously...
+  // Attaching callback on the stream ensures correct timestamp
+  // is captured.
+  cudaLaunchHostFunc(
+      CudaStream(), TimestampCaptureCallback,
+      reinterpret_cast<void*>(&compute_end_ns));
+#endif
 
   // Free BackendMemory used for inputs
   for (BackendMemory* mem : input_memories) {
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -673,12 +673,16 @@ ModelInstanceState::ProcessRequests(
 
   uint64_t compute_end_ns = 0;
 #if defined(TRITON_ENABLE_GPU) && defined(TRITON_ENABLE_STATS)
-  // For GPU, the Execute runs the inference asynchronously...
-  // Attaching callback on the stream ensures correct timestamp
-  // is captured.
-  cudaLaunchHostFunc(
-      c10::cuda::getCurrentCUDAStream().stream(), TimestampCaptureCallback,
-      reinterpret_cast<void*>(&compute_end_ns));
+  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+    // For GPU, the Execute runs the inference asynchronously...
+    // Attaching callback on the stream ensures correct timestamp
+    // is captured.
+    cudaLaunchHostFunc(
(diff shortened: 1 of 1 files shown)


src/libtorch.cc
```diff
@@ -660,7 +672,14 @@ ModelInstanceState::ProcessRequests(
   Execute(&responses, request_count, &input_tensors, &output_tensors);
 
   uint64_t compute_end_ns = 0;
-  SET_TIMESTAMP(compute_end_ns);
+#if defined(TRITON_ENABLE_GPU) && defined(TRITON_ENABLE_STATS)
```
deadeyegoodwin, line 675: Does Execute return before the inference is complete? It doesn't seem so. Is CudaStream() used by pytorch execution engine?

tanmayv25: Yes.. The Execute does not wait for the inference to complete in case of torch::kGPU is specified. Essentially, its an asynchronous operation. See here for details: https://medium.com/@auro_227/timing-your-pytorch-code-fragments-e1a556e81f2
I couldn't find the exact match solution for libtorch. No documentation regarding `torch::cuda::Event`. 

 > It doesn't seem so. Is CudaStream() used by pytorch execution engine?
 I am confused about that too.. Their documentation says torch uses default stream for the current set device.They define it as the stream where most of the compution occurs. They use CUDAStream abstraction on the streams being used by pytorch execution engine. More details here: https://github.com/pytorch/pytorch/blob/master/c10/cuda/CUDAStream.h
 
 I did some empirical experiments, and with reported timing values it looks like stream_ callback is deferred till inference run is complete.

tanmayv25: I guess I can deep -dive into the `c10/cuda/*` files to come with a better solution.

tanmayv25: I can call [`synchronize` ](https://github.com/pytorch/pytorch/blob/master/c10/cuda/CUDAStream.h#L126) on the CUDAStream object returned by pytorch execution engine to ensure `Execute` remains synchronous. But I think this might impact performance.

GuanLuo: But we don't have explicit synchronization in the read output function below, unless libtorch does something fancy like implicit synchronization on first output read.

I am surprised that they would use default stream.. I think the launch host func is somehow working because of the CUDA synchronization behavior related to default stream?

tanmayv25: Yes.. I haven't find it clearly documented, but I believe trying to access `output_tensor` should be internally synchronizing on first output read like : https://github.com/triton-inference-server/pytorch_backend/blob/main/src/libtorch.cc#L867

I have another [commit ](https://github.com/triton-inference-server/pytorch_backend/pull/22/commits/16dc8ef1a45334166b39f91198be0afe7af85437) that ensures the pytorch engine stream is attached the callback..

CoderHam (changes requested):
@tanmayv25 thanks for his fix. If the forward pass is truly async and is only synchronized on reading output from tensors then you are indeed right. While I did confirm that the forward pass should be async and the assumption for it to be synchronized before reading output tensors is valid. 

Additionally I see the overall average latency fell. Which is surprising as generally adding additional synchronizing generally lowers performance.

While multi-gpu perf in pytorch is 'buggy'. Did you test this on a multi-gpu device and see how it impacts perf?

PS: Which model did you use for the tests?

CoderHam (approved):

tanmayv25:
> While I did confirm that the forward pass should be async and the assumption for it to be synchronized before reading output tensors is valid.

So, you confirm the reading first output synchronizes stream? 

> Did you test this on a multi-gpu device and see how it impacts perf?

I haven't yet.. I can try that if you believe this can be an issue..
I am using resnet50_fp32_libtorch model from our perf_model_store.

CoderHam:
Btw cudaLaunchHostFunc was added in cuda 10, should there be a related check around this?

CoderHam:
> So, you confirm the reading first output synchronizes stream?

No, my reading did not confirm that the first o/p synchronizes the stream. I was only able to confirm the forward call being async.

CoderHam:
I guess that makes sense. I was thinking of a case where a user might want to build Triton + pytorch_backend manually for cuda < 10.

deadeyegoodwin:
> 
> 
> I guess that makes sense. I was thinking of a case where a user might want to build Triton + pytorch_backend manually for cuda < 10.

We don't need to directly support that and can leave it to that user to sort out.

tanmayv25:
@deadeyegoodwin  I did some digging and think my latest change should behave as expected or close enough.

The pytorch JIT uses our provided device(CPU/GPU) from [`map_location`](http://man.hubwiz.com/docset/PyTorch.docset/Contents/Resources/Documents/jit.html#torch.jit.load). In case it is CPU the operations are performed synchronously but when the location is GPU the operations are enqueued on a cuda stream. One can use CudaGuard to relay operations through specific streams. In our backend we don't use one hence, it uses a default cuda stream.
@CoderHam  Correct if my understanding appears erroneous.

There is a hiccup in cudaLaunchHostFunc but I think we should be okay.
> Using CUDA 11.3, and 466.11 Driver.
> When Hardware-accelerated GPU scheduling is OFF, cudaLaunchHostFunc serializes (blocks work on all streams).
> When Hardware-accelerated GPU scheduling is ON, cudaLaunchHostFunc does not block work on other streams (correct behavior).

Source: https://forums.developer.nvidia.com/t/does-cudalaunchhostfunc-block-work-added-to-all-streams/170259/19

deadeyegoodwin:
@tanmayv25 If this change is the best we can do then you should go ahead and merge it.
Should we be using CudaGuard in the backend to provide a separate cuda stream for each model instance?

CoderHam:
> @CoderHam Correct if my understanding appears erroneous.
Based on my reading I believe your understanding is correct. If CudaGuard is the preferred way in PyTorch to relay instructions we can consider adding the same. However as I understand the cuda stream used in CopyBuffer is independent from the default stream being used by PyTorch. By using CudaGuard will we be able to (and want to) use a single stream for both?

tanmayv25:
@CoderHam  I think @deadeyegoodwin  meant was to use separate streams for different instances such that the executions can happen on them independently.

PR #23, opened 2021-05-26
Fix timestamp collection to not impact perf

CoderHam:
- Sacrfices true compute output time

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -908,11 +908,16 @@ ModelInstanceState::ReadOutputTensors(
                                         : TRITONSERVER_MEMORY_GPU,
         (device_.type() == torch::kCPU) ? 0 : device_.index());
 
-    // Due to pytorch running inference asynchronously, we need set timestamp
-    // after first output is read to get a more accuracte estimate of compute
-    // infer time. This means the compute output time reported would be smaller
-    // than it is in reality. We do this because synchronizing manually causes a
-    // large drop in performance for fast running models.
+    // PyTorch uses asynchronous execution to run the model. Setting the compute
+    // end timestamp immediately after Execute() does not capture the complete
+    // model execution time. When the first output buffer is accessed/copied by
+    // ProcessTensor(), there is a synchronization that is done to ensure the
+    // data is correctly copied from the output tensor. To avoid overheads of
+    // additional synchronization, we continue to use the default cuda stream.
+    // However the drawback of this is that the compute infer time reported
+    // would be slightly later than it is in reality and the compute output time
+    // reported would be smaller than it is in reality. We allow this because
+    // synchronizing manually negatively impacts performance.
     if (idx == 0) {
       SET_TIMESTAMP(*compute_end_ns);
     }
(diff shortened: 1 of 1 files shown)


src/libtorch.cc
```diff
@@ -929,6 +907,15 @@ ModelInstanceState::ReadOutputTensors(
...
                                         : TRITONSERVER_MEMORY_GPU,
         (device_.type() == torch::kCPU) ? 0 : device_.index());
+
+    // Due to pytorch running inference asynchronously, we need set timestamp
+    // after first output is read to get a more accuracte estimate of compute
+    // infer time. This means the compute output time reported would be smaller
+    // than it is in reality. We do this because synchronizing manually causes a
+    // large drop in performance for fast running models.
+    if (idx == 0) {
+      SET_TIMESTAMP(*compute_end_ns);
```
GuanLuo, line 917: Shouldn't you place this after line 890?
https://github.com/triton-inference-server/pytorch_backend/pull/23/files#diff-859dcc1d94f2d13b0d19e7a14af430cd18eb66395a6f69f2d4886854319c4467L890

src/libtorch.cc
```diff
@@ -929,6 +907,15 @@ ModelInstanceState::ReadOutputTensors(
         (device_.type() == torch::kCPU) ? TRITONSERVER_MEMORY_CPU
                                         : TRITONSERVER_MEMORY_GPU,
         (device_.type() == torch::kCPU) ? 0 : device_.index());
+
```
deadeyegoodwin, line 910: You should be able to move this to line 897 above just after output_buffer pointer is assigned.

deadeyegoodwin (changes requested):

src/libtorch.cc
CoderHam: I did try that first. I shared number for the same in the ticket. It appears the synchronization occurs during the responder.ProcessTensor call.

GuanLuo: Why would that be the case? Are you saying that the content in the output_buffer is not valid until within the responder? But the responder doesn't interact with the torch library at all, how does the synchronization happen?

CoderHam: I am not 100% clear but this is what I believe is happening:
The output_buffer is a pointer to memory on GPU that the output tensor from pytorch will be stored. While trying to access data at the location, pytorch synchronizes and populates the output_buffer. While this is odd it does help with performance.
I could not find any documentation from pytorch that explains this behavior.

deadeyegoodwin: There's no way for pytorch to do that. Once we have a char* pointer to the data, accessing it via that pointer will not trigger anything that pytorch can be aware of (if there was some OS level page protection on the memory there could be some OS action triggered... but not in the pytorch code).

CoderHam: I can run some more tests to understand what causes this behavior.

deadeyegoodwin (approved):

GuanLuo (approved):

PR #26, opened 2021-06-07
Do not throw error when DISABLE_OPTIMIZED_EXECUTION is missing.

CoderHam:
- parameters is present but empty in the model config json if not specified in config.prototxt
Cause of failures in master CI

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -195,13 +195,20 @@ ModelState::ParseParameters()
   triton::common::TritonJson::Value params;
   bool status = model_config_.Find("parameters", &params);
   if (status) {
-    RETURN_IF_ERROR(ParseParameter(
-        params, "DISABLE_OPTIMIZED_EXECUTION", &disable_optimized_execution_));
-    LOG_MESSAGE(
-        TRITONSERVER_LOG_INFO,
-        (std::string("Optimized execution is ") +
-         (disable_optimized_execution_ ? "disabled" : "enabled"))
-            .c_str());
+    TRITONSERVER_Error* err = ParseParameter(
+        params, "DISABLE_OPTIMIZED_EXECUTION", &disable_optimized_execution_);
+    if (err != nullptr) {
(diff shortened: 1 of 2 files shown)


tanmayv25 (approved):

src/libtorch.cc
```diff
@@ -195,13 +195,20 @@ ModelState::ParseParameters()
...
-        (std::string("Optimized execution is ") +
-         (disable_optimized_execution_ ? "disabled" : "enabled"))
-            .c_str());
+    TRITONSERVER_Error* err = ParseParameter(
+        params, "DISABLE_OPTIMIZED_EXECUTION", &disable_optimized_execution_);
+    if (err != nullptr) {
+      if (TRITONSERVER_ErrorCode(err) != TRITONSERVER_ERROR_NOT_FOUND) {
+        return err;
+      } else {
+        LOG_MESSAGE(
```
deadeyegoodwin, line 204: Why are you printing a message only when TRITONSERVER_ERROR_NOT_FOUND? Don't you just want to set disable_optimized_execution_ to a default value (false) when TRITONSERVER_ERROR_NOT_FOUND is returned. And then always log the message?

deadeyegoodwin (changes requested):

src/libtorch.cc
CoderHam: Yes we would want to log the message always.
The default value for disable_optimized_execution_ is already set in the constructor.

deadeyegoodwin: But you should not assume that disable_optimized_execution_ is unchanged when the function returns TRITONSERVER_ERROR_NOT_FOUND. Because of that and to make the code easier to understand you should make sure ParsseParameter is documented (and implemented) to not change the value for NOT_FOUND and also put a note in your code where you call it.

CoderHam: That is correct. ParseParameters is designed to check if the key is present and respond with an error is it is absent. If this step fails then the boolean value is not updated.
I will update the documentation of ParseParameters and add a note before the corresponding call.

tanmayv25 (approved):

deadeyegoodwin (approved):

PR #28, opened 2021-06-10
Return appropriate error if the output is a scalar

CoderHam:
Error reported is of the form `output 'output__0' is a scalar which is not supported.`

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -942,6 +942,16 @@ ModelInstanceState::ReadOutputTensors(
       batchn_shape.push_back(*itr);
     }
 
+    if (batchn_shape.size() == 0) {
+      RESPOND_ALL_AND_RETURN_IF_ERROR(
+          responses, request_count,
+          TRITONSERVER_ErrorNew(
+              TRITONSERVER_ERROR_INVALID_ARG,
+              (std::string("output '") + name +
+               "' is a scalar which is not supported.")
+                  .c_str()));
+    }
+
     responder.ProcessTensor(
         name, output_dtype, batchn_shape, output_buffer,
         (device_.type() == torch::kCPU) ? TRITONSERVER_MEMORY_CPU


CoderHam:
Resolves segmentation fault that was reported by https://github.com/triton-inference-server/server/issues/2985

deadeyegoodwin (approved):
Do we have a test for this?

src/libtorch.cc
```diff
@@ -942,6 +942,16 @@ ModelInstanceState::ReadOutputTensors(
...
 
+    if (batchn_shape.size() == 0) {
+      RESPOND_ALL_AND_RETURN_IF_ERROR(
+          responses, request_count,
+          TRITONSERVER_ErrorNew(
+              TRITONSERVER_ERROR_INVALID_ARG,
+              (std::string("output '") + name +
+               "' is a scalar which is not supported.")
+                  .c_str()));
+    }
```
GuanLuo, line 953: Can we just prepare it as output tensor with shape [1]?

CoderHam:
> Do we have a test for this?

We don't. I can add a test with the fasterrcnn model the user provided me.

src/libtorch.cc
CoderHam: While that would solve the issue, it would not be true to the actual output from the model.

Are we willing to sacrifice on the accuracy of output shape and allow scalar outputs by doing so? 
@deadeyegoodwin  @tanmayv25 I'd appreciate both of your opinion on the matter. 

On one hand we can avoid asking users to trace their model. If other backends allow scalar outputs do we extend this behaviour to them as well and document the same? 

If we do this on the output side to allow supporting scalars what stops us from doing the same for input scalars?

deadeyegoodwin:
Generally, we could automatically map a [ 1 ] vector in the model configuration to a scalar if that is what the model requires. It would actually be nice to do this for all our framework backends . We even have a poorly title ticket for it: https://jirasw.nvidia.com/browse/DLIS-675

Hemant why don't you add a link to this PR along with the suggestion that we could instead handle scalars better. I think we should address this limitation across all backends if/when we feel that is a priority.

src/libtorch.cc
```diff
@@ -942,6 +942,16 @@ ModelInstanceState::ReadOutputTensors(
...
+
     responder.ProcessTensor(
         name, output_dtype, batchn_shape, output_buffer,
```
GuanLuo, line 956: By the way, similar to this issue, since there is no way to validate the output shape until the model is executed, when the model "supports" batching, you should verify the first dimension of the output matches the input.

src/libtorch.cc
tanmayv25: I too believe if we can support the execution of scalar tensors by automatically reshaping the tensors then we should definitely do so. 
A user who doesn't have control over the model will surely appreciate it. Actually, they should not care about whether Triton is treating scalar inputs as [1] tensors as long as they can load the model and make an inference call to it. 
They can use reshape() in our model_config to support scalars on input size. 

However, the behavior should be consistent across all the backends. As we don't have a proper support for scalar tensors. I am fine returning this error for now.

src/libtorch.cc
CoderHam: Nice chat! I'd like to address that in a follow up ticket+PR though since it will need a custom model to test the same.

deadeyegoodwin: Yes, this change should go as is, we can use the existing ticket for more general scalar support

PR #30, opened 2021-07-06
Fix typo in NGC url

CoderHam:


CoderHam:
cc @dzier to make sure we don't miss this during code freeze

PR #32, opened 2021-07-19
Document effects of libtorch asynchronous execution on backend

tanmayv25:
Documenting current behavior for now. Should be updated with tritonserver enhancement to detect cuda context corruption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,3 +120,14 @@ key: "DISABLE_OPTIMIZED_EXECUTION"
     }
 }
 ```
+
+### Important Note
+
+The execution of pytorch model on GPU is asynchronous in nature. See
+[here](https://pytorch.org/docs/stable/notes/cuda.html#asynchronous-execution) for more details.
+Consequently, an error in pytorch model execution may be raised in the next few inference requests to the
+server. Setting environment variable `CUDA_LAUNCH_BLOCKING=1` when launching server will help in correctly
+debugging failing cases by forcing synchronous execution.
+
+The PyTorch model in such cases may or may not recover from the failed state and a restart of application
+may be required to continue serving successfully.
--- a/README.md
+++ b/README.md
@@ -123,11 +123,12 @@ key: "DISABLE_OPTIMIZED_EXECUTION"
 
 ### Important Note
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,3 +120,14 @@ key: "DISABLE_OPTIMIZED_EXECUTION"
     }
 }
 ```
+
+### Important Note
+
+The execution of pytorch model on GPU is asynchronous in nature. See
+[here](https://pytorch.org/docs/stable/notes/cuda.html#asynchronous-execution) for more details.
+Consequently, an error in pytorch model execution may be raised in the next few inference requests to the
```
CoderHam, line 128: raised _during_ the next ...?

README.md
```diff
@@ -120,3 +120,14 @@ key: "DISABLE_OPTIMIZED_EXECUTION"
...
+server. Setting environment variable `CUDA_LAUNCH_BLOCKING=1` when launching server will help in correctly
+debugging failing cases by forcing synchronous execution.
+
+The PyTorch model in such cases may or may not recover from the failed state and a restart of application
```
CoderHam, line 132: restart of **the server**

CoderHam (changes requested):

CoderHam:
@deadeyegoodwin we should also call our the multiple instance perf issues with the pytorch backend due to thread specific caching in pytorch. Do you think it would be a good idea to have this here?

deadeyegoodwin:
> 
> 
> @deadeyegoodwin we should also call our the multiple instance perf issues with the pytorch backend due to thread specific caching in pytorch. Do you think it would be a good idea to have this here?

Yes, it would be good to refer to that open issue as well.

CoderHam:
@tanmayv25 do you want to add in the content in or prefer I do that in a follow up PR?
PS: PyTorch issue https://github.com/pytorch/pytorch/issues/27902

tanmayv25:
@CoderHam  I think you would be able to better document it with a follow-up PR.. I will bullet these points so we can point to various issues observed with the framework.

CoderHam:
Sounds good :)

CoderHam (approved):

CoderHam (approved):

PR #33, opened 2021-07-26
Use InferenceMode guard instead of NoGradGuard

CoderHam:
- Support for InferenceMode was added in Pytorch 1.9.0

Read more here: https://pytorch.org/cppdocs/notes/inference_mode.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <!--
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
--- a/README.md
+++ b/README.md
@@ -99,7 +99,8 @@ $ make install
 
 ### Parameters
 
-Disabling the optimized execution of the PyTorch models is done through the Parameters section of the model's 'config.pbtxt' file.
+Triton exposes some flags to control the execution mode of the TorchScript models through
+the Parameters section of the model's 'config.pbtxt' file.
 
 * `DISABLE_OPTIMIZED_EXECUTION`: Boolean flag to disable the optimized execution
 of TorchScript models. By default the optimized execuiton is always enabled. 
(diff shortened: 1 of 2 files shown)


tanmayv25 (changes requested):

tanmayv25:
Can you share the performance improvement you see with the switch?  `resnet50` may be?

CoderHam:
> Can you share the performance improvement you see with the switch? `resnet50` may be?

Yes I am gathering some perf numbers for resnet50 batch size 8 and 128 before and after these changes. I will share them shortly

src/libtorch.cc
```diff
@@ -786,7 +789,8 @@ ModelInstanceState::Execute(
     torch::jit::setGraphExecutorOptimize(
         !model_state_->DisabledOptimizedExecution());
 
-    torch::NoGradGuard no_grad;
+    // InferenceMode is a faster and more restricive version of NoGradGuard
```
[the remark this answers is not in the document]
CoderHam: Done

tanmayv25 (dismissed):

deadeyegoodwin:
Two things to think about. resnet50 is not representative of all models. It seems like InferenceMode is the official, recommended pytorch way for inferencing? If that is so then there may be ongoing enhancements that will improve its functioning over time. So it seems like having it as an option seems safer than not having it at all.

src/libtorch.cc
```diff
@@ -786,7 +811,8 @@ ModelInstanceState::Execute(
     torch::jit::setGraphExecutorOptimize(
         !model_state_->DisabledOptimizedExecution());
 
-    torch::NoGradGuard no_grad;
+    // enable/disable inference mode
+    torch::InferenceMode infer_guard(model_state_->InferenceMode());
```
tanmayv25, line 815: Shouldn't we fallback to NoGradGuard if INFERENCE_MODE is set false. I mean that is the current default behavior. 
According to the numbers you shared it looked like in some cases NoGradGuard was giving the best performance.

CoderHam: You are right. Fixing this now.

tanmayv25 (approved):

tanmayv25 (dismissed):

tanmayv25 (approved):

src/libtorch.cc
GuanLuo: So if infer model is set, the no_grad below will be ignored?

README.md
```diff
@@ -121,6 +122,28 @@ key: "DISABLE_OPTIMIZED_EXECUTION"
...
+this mode gets better performance by disabling autograd.
+
+Please note that in some models, InferenceMode might not benefit performance
+and in fewer cases might impact performance negatively.
+
+The section of model config file specifying this parameters will look like:
+
+```
+parameters: {
+key: "INFERENCE_MODE"
```
GuanLuo, line 140: If inference mode is going to be the official way for PyTorch inferencing, shouldn't it be set by default unless user want the model to be run in NoGradMode?

CoderHam: I was going to make this the default to keep to the behavior in the python and C++ executions users must enable the same. 

In the future if we see that the InferenceMode consistently improves performance we can enable it by default. For now I would prefer it be an opt in flag instead of opt out.

src/libtorch.cc
CoderHam: Yes. When enabled, InferenceMode enables NoGrad. Technically NoGrad is a subset of InferenceMode

README.md
deadeyegoodwin: I agree it should be off by default. In general we should prefer to maintain backwards compat even in perf until we have high confidence that the new mode is better.

deadeyegoodwin (approved):

PR #36, opened 2021-09-16
Add verbose error for when output is a list / dict of tensors or scalar

CoderHam:


src/libtorch.cc
```diff
@@ -823,8 +823,16 @@ ModelInstanceState::Execute(
...
     } else {
-      auto model_output_tensor = model_outputs_.toTensor();
-      output_tensors->push_back(model_output_tensor);
+      try {
+        auto model_output_tensor = model_outputs_.toTensor();
+        output_tensors->push_back(model_output_tensor);
+      }
+      catch (std::exception& exx) {
+        throw std::invalid_argument(
+            "Output of torch model should be tensor or a tuple of tensors, not "
```
deadeyegoodwin, line 832: I see above that we support a tuple of tensors. How much work would it be to also support list and dictionary?

deadeyegoodwin (approved):

src/libtorch.cc
CoderHam: Support for tuple of tensors is relatively straightforward but we can add processing for list and dictionaries as well. 
PS: If there is 1 o/p it is a tensor, if the model has N outputs, pytorch returns it as a tuple of tensors. 

We can convert the list of tensors to N tensors i.e. if we have a model that outputs List[Tensor]) where the List is of variable size, we can convert this to (Tensor, Tensor ....) but how do we support this in our model config?

For the case where the List / Dict is of definite length N, we can just convert the List / Dict to N Tensors (for dict how do we maintain order?) and support it in that way.

CoderHam: @deadeyegoodwin we have an additional case that is very common. FasterRCNN returns a List[Dict[Tensors]] and takes an input of List[Tensors]. Even if we add support for outputs in the form of List / Dict of Tensors, how do we support such cases?

PR #40, opened 2021-10-06
Enable nvfuser by default

CoderHam:
- Add option to disable nvfuser

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -80,6 +80,7 @@ class ModelState : public BackendModel {
 
   bool DisabledOptimizedExecution() { return disable_optimized_execution_; }
   bool InferenceMode() { return inference_mode_; }
+  bool DisabledNvfuser() { return disable_nvfuser_; }
 
  private:
   ModelState(TRITONBACKEND_Model* triton_model);
@@ -93,6 +94,9 @@ class ModelState : public BackendModel {
 
   // Flag to indicate whether inference mode is enabled
   bool inference_mode_;
+
+  // Flag to indicate whether nvfuser is disabled
+  bool disable_nvfuser_;
 };
 
 
@@ -133,7 +137,7 @@ ModelState::Create(TRITONBACKEND_Model* triton_model, ModelState** state)
 
 ModelState::ModelState(TRITONBACKEND_Model* triton_model)
     : BackendModel(triton_model), disable_optimized_execution_(false),
-      inference_mode_(false)
+      inference_mode_(false), disable_nvfuser_(false)
 {
 }
 
@@ -170,6 +174,14 @@ ModelState::LoadModel(
   // InferenceMode should be used to guard all tensors operations including
   // model loading: https://pytorch.org/cppdocs/notes/inference_mode.html
   torch::InferenceMode infer_guard(InferenceMode());
+
+  if (!DisabledNvfuser()) {
+    torch::jit::overrideCanFuseOnCPU(false);
+    torch::jit::overrideCanFuseOnGPU(false);
+    torch::jit::setTensorExprFuserEnabled(false);
+    torch::jit::RegisterCudaFuseGraph::registerPass(true);
+  }
+
   try {
     std::istringstream model_stream(model_data_str);
     torch_model->reset(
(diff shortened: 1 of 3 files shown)


src/libtorch.cc
```diff
@@ -88,11 +89,14 @@ class ModelState : public BackendModel {
   // Parses and validates parameters in config
   TRITONSERVER_Error* ParseParameters();
 
-  // Flag to indicate whether optimized execution is disabled
+  // Flag to indicate whether optimized execution is disabled. Defaults to false.
   bool disable_optimized_execution_;
 
-  // Flag to indicate whether inference mode is enabled
+  // Flag to indicate whether inference mode is enabled. Defaults to false.
   bool inference_mode_;
```
deadeyegoodwin, line 96: Having enables seems like it would be more typical than having disables. But should be consistent across all 3. And name them all consistently as enable_* or disable_*

src/libtorch.cc
```diff
@@ -237,6 +249,24 @@ ModelState::ParseParameters()
         TRITONSERVER_LOG_INFO, (std::string("Inference Mode is ") +
                                 (inference_mode_ ? "enabled" : "disabled"))
                                    .c_str());
+
+    // If 'DISABLE_NVFUSER' is not present in 'parameters' then no
+    // update is made to 'disable_nvfuser_'.
+    err = ParseParameter(
+        params, "DISABLE_NVFUSER", &disable_nvfuser_);
```
deadeyegoodwin, line 256: ENABLE_NVFUSER?

deadeyegoodwin (changes requested):

src/libtorch.cc
CoderHam: Since the default value for these 3 flags are different do we still want to name them consistently? 
optimized_execution is on by default.
inference_mode is off by default.
nvfuser is off by default.

deadeyegoodwin: More important to have consistent naming then to match the defaults.

CoderHam: Fixed

src/libtorch.cc
```diff
@@ -169,7 +173,16 @@ ModelState::LoadModel(
 
   // InferenceMode should be used to guard all tensors operations including
   // model loading: https://pytorch.org/cppdocs/notes/inference_mode.html
-  torch::InferenceMode infer_guard(InferenceMode());
+  torch::InferenceMode infer_guard(EnabledInferenceMode());
+
+  // NvFuser is only supported for GPU instances
```
deadeyegoodwin, line 178: When nvfuser becomes the default in pytorch, setting ENABLE_NVFUSER to false will not disable it. I think we need to have a 3-way setting for it (could use std::optional) so that we can know if it was explicitly set to true or false or not set at all. Then we can explicitly enable or disable or do nothing. PyTorch team mentioned that they are working with upstream to make it the default and it seems like being able to explicitly have it on or off may be important.

deadeyegoodwin (changes requested):

src/libtorch.cc
```diff
@@ -78,8 +78,9 @@ class ModelState : public BackendModel {
       std::string* model_path,
       std::unique_ptr<torch::jit::script::Module>* torch_model);
 
-  bool DisabledOptimizedExecution() { return disable_optimized_execution_; }
-  bool InferenceMode() { return inference_mode_; }
+  bool EnabledOptimizedExecution() { return enable_optimized_execution_; }
+  bool EnabledInferenceMode() { return enable_inference_mode_; }
+  std::pair<bool, bool> EnabledNvfuserPair() { return enable_nvfuser_pair_; }
```
deadeyegoodwin, line 83: Return as const&

deadeyegoodwin (changes requested):

README.md
```diff
@@ -144,6 +144,24 @@ key: "INFERENCE_MODE"
 }
 ```
 
+* `ENABLE_NVFUSER`: Boolean flag to enable the NvFuser (CUDA Graph Fuser) optimization for
+TorchScript models. By default the NvFuser is disabled. 
+
+Please note that in some models generated using trace in old PyTorch versions might not work
+correctly with NvFuser. We recommend using scripting and a recently version of PyTorch
```
tanmayv25, line 151: *recent version

src/libtorch.cc
```diff
@@ -810,10 +844,27 @@ ModelInstanceState::Execute(
...
     torch::jit::setGraphExecutorOptimize(
-        !model_state_->DisabledOptimizedExecution());
+        model_state_->EnabledOptimizedExecution());
 
     // enable/disable inference mode - supersedes NoGradGuard
-    torch::InferenceMode infer_guard(model_state_->InferenceMode());
+    torch::InferenceMode infer_guard(model_state_->EnabledInferenceMode());
+
+    // NvFuser is only supported for GPU instances. No change is made if
+    // enable_nvfuser is not set by user
```
tanmayv25, line 853: Can you add a link to how you came up with the following truth vales for different options below?

CoderHam: Unfortunately, there is no user facing doc for the C++ API calls for NvFuser, I found it by digging through https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/python/init.cpp and by discussing with some pytorch framework experts.

deadeyegoodwin (approved):

tanmayv25 (approved):

PR #42, opened 2021-10-29
add options for certain optimizations

ryanleary:
Some models seem to perform poorly with the default set of JIT optimizations. Expose some of these API parameters in the config so that these models can be run in Triton.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,8 @@ key: "INFERENCE_MODE"
 ```
 
 * `ENABLE_NVFUSER`: Boolean flag to enable the NvFuser (CUDA Graph Fuser) optimization for
-TorchScript models. If not specified, the default pytorch fuser is used. 
+TorchScript models. If not specified, the default pytorch fuser is used.
+If `ENABLE_TENSOR_FUSER` is specified and _disabled_, the NvFuser configuration is ignored.
 
 Please note that in some models generated using trace in old PyTorch versions might not work
 correctly with NvFuser. We recommend using scripting and a recent version of PyTorch
@@ -170,7 +171,7 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
 
     `ENABLE_JIT_PROFILING`
 
-    `ENABLE_TORCH_TENSOR_FUSER`
+    `ENABLE_TENSOR_FUSER`
 
 ### Important Note
 
--- a/README.md
+++ b/README.md
@@ -162,6 +162,16 @@ key: "ENABLE_NVFUSER"
 }
 ```
 
+* Additional Optimizations: Three additional boolean parameters are available to disable
+certain Torch optimizations that can sometimes cause latency regressions in models with
+complex execution modes and dynamic shapes. If not specified, all are enabled by default.
+
+    `ENABLE_JIT_EXECUTOR`
+
+    `ENABLE_JIT_PROFILING`
+
+    `ENABLE_TORCH_TENSOR_FUSER`
+
 ### Important Note
 
 * The execution of pytorch model on GPU is asynchronous in nature. See
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -26,7 +26,6 @@
 
 #include <stdint.h>
 #include <exception>
-
 #include "libtorch_utils.h"
 #include "triton/backend/backend_common.h"
 #include "triton/backend/backend_input_collector.h"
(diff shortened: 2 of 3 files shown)


src/libtorch.cc
```diff
@@ -848,6 +917,10 @@ ModelInstanceState::Execute(
     // enable/disable optimized execution
     torch::jit::setGraphExecutorOptimize(
         model_state_->EnabledOptimizedExecution());
+    torch::jit::setTensorExprFuserEnabled(
+        model_state_->EnabledTensorExprFuser());
```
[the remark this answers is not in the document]
CoderHam: > So I think this fuser logic needs to be combined with the nvfuser_pair logic below so that is fuser is false both fuser and nvfuser is disabled, otherwise use the existing logic below. @CoderHam what do you think?

Yes. It does need to be combined to ensure consistent behaviour. 

@ryanleary why do we need ENABLE_TENSOR_FUSER as well as ENABLE_NVFUSER?

README.md
```diff
@@ -162,6 +163,16 @@ key: "ENABLE_NVFUSER"
 }
 ```
 
+* Additional Optimizations: Three additional boolean parameters are available to disable
+certain Torch optimizations that can sometimes cause latency regressions in models with
+complex execution modes and dynamic shapes. If not specified, all are enabled by default.
+
+    `ENABLE_JIT_EXECUTOR`
```
CoderHam, line 170: Use bullets / numbers here?

src/libtorch.cc
```diff
@@ -26,6 +26,7 @@
 
 #include <stdint.h>
 #include <exception>
+
```
CoderHam, line 29: Extra empty line not needed

src/libtorch.cc
```diff
@@ -252,27 +264,85 @@ ModelState::ParseParameters()
          " for model instance '" + Name() + "'")
             .c_str());
 
-    // If 'ENABLE_NVFUSER' is not present in 'parameters' then no
-    // update is made to 'enable_nvfuser'.
-    bool enable_nvfuser = false;
-    err = ParseParameter(params, "ENABLE_NVFUSER", &enable_nvfuser);
+    // If 'ENABLE_TORCH_TENSOR_FUSER' is not present in 'parameters' then no
```
CoderHam, line 267: ENABLE_TORCH_TENSOR_FUSER -> ENABLE_TENSOR_FUSER

src/libtorch.cc
```diff
@@ -252,27 +264,85 @@ ModelState::ParseParameters()
...
+    if (err != nullptr) {
+      if (TRITONSERVER_ErrorCode(err) != TRITONSERVER_ERROR_NOT_FOUND) {
+        return err;
+      } else {
+        TRITONSERVER_ErrorDelete(err);
+      }
+    } else {
+      LOG_MESSAGE(
+          TRITONSERVER_LOG_INFO,
+          (std::string("Torch tensor fuser is ") +
```
CoderHam, line 280: Just "Tensor fuser"? since we do not use Torch elsewhere.

src/libtorch.cc
```diff
@@ -98,6 +102,12 @@ class ModelState : public BackendModel {
   // Flag to indicate whether inference mode is enabled. Defaults to false.
   bool enable_inference_mode_;
 
+  // Flag to enable/disable various JIT components. Defaults to true.
+  // Tensor fuser configuration will be overridden by configuration of nvfuser.
+  bool enable_tensor_fuser_;
+  bool enable_jit_profiling_;
```
CoderHam, line 108: Do we want enable_jit_profiling_  to default to true? Would it not cause perf impacts?

src/libtorch.cc
```diff
@@ -79,6 +80,9 @@ class ModelState : public BackendModel {
       std::unique_ptr<torch::jit::script::Module>* torch_model);
 
   bool EnabledOptimizedExecution() { return enable_optimized_execution_; }
+  bool EnabledTensorExprFuser() { return enable_tensor_fuser_; }
+  bool EnabledJitProfilingMode() { return enable_jit_profiling_; }
```
CoderHam, line 84: A mode symbolizes it will have a non-binary value but in this case it is a flag instead. Shouldn't `EnabledJitProfiling()` be enough?
Similarly for EnabledJitExecutorMode.

src/libtorch.cc
```diff
@@ -848,6 +918,10 @@ ModelInstanceState::Execute(
     // enable/disable optimized execution
     torch::jit::setGraphExecutorOptimize(
         model_state_->EnabledOptimizedExecution());
+    torch::jit::setTensorExprFuserEnabled(
+        model_state_->EnabledTensorExprFuser());
+    torch::jit::getProfilingMode() = model_state_->EnabledJitProfilingMode();
+    torch::jit::getExecutorMode() = model_state_->EnabledJitExecutorMode();
```
CoderHam, line 924: What is the significance of Executor Mode?

src/libtorch.cc
```diff
@@ -947,8 +1021,8 @@ ModelInstanceState::SetInputTensors(
     // The input must be in contiguous CPU/GPU memory.
     std::vector<std::pair<TRITONSERVER_MemoryType, int64_t>> alloc_perference;
     if (device_.is_cpu()) {
-      alloc_perference = {{TRITONSERVER_MEMORY_CPU_PINNED, 0},
-                          {TRITONSERVER_MEMORY_CPU, 0}};
+      alloc_perference = {
```
CoderHam, line 1024: Can you run the format.py from the common repo? More instructions here:
https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#coding-convention

CoderHam (changes requested):

deadeyegoodwin:
@CoderHam I think we are going to want to handle these new parameters in the same way as nvfuser so that we don't do anything unless the user explicitly requests it (wither enable or disable). I don't have confidence in setting pytorch options by default so if the user doesn't explicitly specify it I don't think we should call them. @ryanleary  is going to close this and create a branch directly in the repo so it will be easier for us to contribute to this PR.

ryanleary:
Closing per discussion. Tried to address most remaining feedback.

PR #44, opened 2021-11-14
TorchTRT support

borisfom:


CMakeLists.txt
```diff
@@ -41,11 +41,20 @@ project(tritonpytorchbackend LANGUAGES C CXX)
 #   - Set TRITON_PYTORCH_DOCKER_IMAGE to use the docker image of
 #     PyTorch to base the build off.
 #
+# TRITON_PYTORCH_ENABLE_TORCHTRT: enables Torch-TRT runtime support.
+#   With TRITON_PYTORCH_DOCKER_IMAGE: set if to OFF for images before 21.11
+#   Without TRITON_PYTORCH_DOCKER_IMAGE:
+#     If you leave it ON: set TRITON_PYTORCH_TORCHTRT_LIB_PATH to Torch-TRT runtime lib.
```
deadeyegoodwin, line 47: @CoderHam seems like the jetson build will need some updates for this.

deadeyegoodwin (commented):
Looks ok to me, but @CoderHam  should review as well.

CMakeLists.txt
CoderHam: The general recommendation for people building a specific version of Triton is to use the corresponding release branch. I would suggest not adding sections like 
```
#   With TRITON_PYTORCH_DOCKER_IMAGE: set if to OFF for images before 21.11
```
Similarly for TRITON_PYTORCH_TORCHTRT_LIB_PATH, we want to follow a part like we do for TRITON_PYTORCH_LIB_PATHS. In the case where the TRITON_PYTORCH_LIB_PATHS is specified, we set TRITON_PYTORCH_DOCKER_BUILD to ON.

When TRITON_PYTORCH_TORCHTRT_LIB_PATH is specified we should mandate the specification of TRITON_PYTORCH_LIB_PATHS and turn Docker build to be ON.

CoderHam: Coming back to Jetson build, yes we will need to pass TRITON_PYTORCH_TORCHTRT_LIB_PATH to build.py in our jetson_build.sh script.

Currently we use a single arg called for library paths to pass these paths to build.py 

My suggestion is to keep a single argument for TRITON_PYTORCH_LIB_PATHS where the paths are specified in a `;` separated fashion as is the norm.

CMakeLists.txt
```diff
@@ -41,11 +41,20 @@ project(tritonpytorchbackend LANGUAGES C CXX)
...
 #
+# TRITON_PYTORCH_ENABLE_TORCHTRT: enables Torch-TRT runtime support.
+#   With TRITON_PYTORCH_DOCKER_IMAGE: set if to OFF for images before 21.11
+#   Without TRITON_PYTORCH_DOCKER_IMAGE:
+#     If you leave it ON: set TRITON_PYTORCH_TORCHTRT_LIB_PATH to Torch-TRT runtime lib. 
+#
+
 option(TRITON_ENABLE_GPU "Enable GPU support in backend" ON)
 option(TRITON_ENABLE_STATS "Include statistics collections in backend" ON)
+option(TRITON_PYTORCH_ENABLE_TORCHTRT "Enable TorchTRT support" ON)
```
CoderHam, line 52: Add a check for TRITON_PYTORCH_ENABLE_TORCHTRT so that TRITON_PYTORCH_ENABLE_TORCHTRT  is not ON when TRITON_ENABLE_GPU is set to OFF

CMakeLists.txt
```diff
@@ -147,12 +167,13 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
       "libopencv_core.so"
       "libpng16.so"
   )
-
+
```
CoderHam, line 170: Remove extra spaces

CMakeLists.txt
```diff
@@ -136,6 +155,7 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
       "libtorch.so"
       "libtorch_cpu.so"
       "libtorch_cuda.so"
+      "libtorch_global_deps.so"
```
CoderHam, line 158: Is this needed if TRITON_PYTORCH_ENABLE_TORCHTRT is OFF?

CMakeLists.txt
```diff
@@ -167,7 +188,10 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
     COMMAND docker cp pytorch_backend_ptlib:/opt/conda/lib/python3.8/site-packages/torch/lib/libtorch.so libtorch.so
     COMMAND docker cp pytorch_backend_ptlib:/opt/conda/lib/python3.8/site-packages/torch/lib/libtorch_cpu.so libtorch_cpu.so
     COMMAND docker cp pytorch_backend_ptlib:/opt/conda/lib/python3.8/site-packages/torch/lib/libtorch_cuda.so libtorch_cuda.so
+    COMMAND docker cp pytorch_backend_ptlib:/opt/conda/lib/python3.8/site-packages/torch/lib/libtorch_global_deps.so libtorch_global_deps.so
```
CoderHam, line 191: Same, needed if TRITON_PYTORCH_ENABLE_TORCHTRT is OFF?

CMakeLists.txt
```diff
@@ -107,6 +116,16 @@ endif() # TRITON_ENABLE_GPU
 #
 configure_file(src/libtriton_pytorch.ldscript libtriton_pytorch.ldscript COPYONLY)
 
+if (TRITON_PYTORCH_ENABLE_TORCHTRT)
+   set(TORCHTRT_COPY_COMMAND docker cp pytorch_backend_ptlib:/opt/conda/lib/python3.8/site-packages/torch_tensorrt/lib/libtorchtrt_runtime.so libtorchtrt_runtime.so)
```
CoderHam, line 120: This block needs to be surrounded by TRITON_PYTORCH_DOCKER_BUILD. Why not move to under 129?

CMakeLists.txt
```diff
@@ -332,7 +357,12 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
     DESTINATION ${CMAKE_INSTALL_PREFIX}/backends/pytorch
   )
 
-  FOREACH(plib ${PT_LIBS} ${OPENCV_LIBS})
+  install(
+    FILES       ${CMAKE_CURRENT_BINARY_DIR}/torchtrtc
```
CoderHam, line 361: Run formatter. Extra space is not needed

CMakeLists.txt
```diff
@@ -332,7 +357,12 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
...
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/bin
```
CoderHam, line 362: We should have this as `${CMAKE_INSTALL_PREFIX}/backends/pytorch` as well

CoderHam (changes requested):
@borisfom thank you for your contribution.

CMakeLists.txt
borisfom: Yes I agree single TRITON_PYTORCH_LIB_PATHS is the best way to go here.

CMakeLists.txt
borisfom: Well it seems to work without it - on the other hand, that may be just we never encountered the issue. I believe it's safer to always having it linked in.

CMakeLists.txt
borisfom: Well, we do not even care to disable CUDA libs copy if CUDA is not enabled - this should do no harm.

CMakeLists.txt
borisfom: Other macros set in this block are not docker-specific. I can split the block in two if that matters.

CMakeLists.txt
borisfom: How do I do that ?

CMakeLists.txt
borisfom: It also seems, for integration purposes, I should set default for TRITON_PYTORCH_ENABLE_TORCHTRT to OFF. Maybe later we can flip it.

borisfom: I don't see why Jetson would need to handle that differently - PyTorch docker image for Jetson will have TorchTRT in.

CoderHam: @deadeyegoodwin do we plan to enable TRITON_PYTORCH_ENABLE_TORCHTRT by default in 20.11?

CoderHam: > I don't see why Jetson would need to handle that differently - PyTorch docker image for Jetson will have TorchTRT in.

For jetson, we do not use docker for our build. Currently we do not officially support Pytorch on Jetson but when we do it will be require us to copy the respective libraries / source code to a local dir and link against them instead of using docker like we do on x86.

In the long term future we might want to dockerize our Jetson build but till then we are limited as above.

CMakeLists.txt
CoderHam: https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#coding-convention elaborates how to install the necessary deps and run the formatter.

CMakeLists.txt
CoderHam: Yes that would be ideal since we would anyway need it to function as such for when we configure it for the non-docker based Jetson build.

CMakeLists.txt
CoderHam: Yes, you have a point.

CMakeLists.txt
CoderHam: Out of curiosity, what is the issue you encounter if libtorch_global_deps.so is not linked?

CMakeLists.txt
deadeyegoodwin: > 
> 
> @deadeyegoodwin do we plan to enable TRITON_PYTORCH_ENABLE_TORCHTRT by default in 20.11?

21.12? 21.11 is done. We can make it default in 21.12 if we have sufficient testing. Is there a reason not to? We can turn it off for jetson if that is necessary.

CMakeLists.txt
```diff
@@ -100,8 +108,22 @@ FetchContent_MakeAvailable(repo-common repo-core repo-backend)
 #
 if(${TRITON_ENABLE_GPU})
   find_package(CUDAToolkit REQUIRED)
+else()
```
CoderHam, line 111: Instead of turning it off you should throw an error
See how to do this in the ORT backend: https://github.com/triton-inference-server/onnxruntime_backend/blob/main/CMakeLists.txt#L101-L105

CMakeLists.txt
```diff
@@ -100,8 +108,22 @@ FetchContent_MakeAvailable(repo-common repo-core repo-backend)
...
+	"-ltorch"
+	"-ltorchvision"
+)
+if (TRITON_PYTORCH_ENABLE_TORCHTRT)
+   message(STATUS "Torch-TRT support is ON")
+   set(TRITON_PYTORCH_LIBS
+       ${TRITON_PYTORCH_LIBS}
+       "-ltorchtrt_runtime"
+   )
+endif() #TRITON_PYTORCH_ENABLE_TORCHTRT
```
CoderHam, line 125: space after `#`

CMakeLists.txt
```diff
@@ -41,8 +41,16 @@ project(tritonpytorchbackend LANGUAGES C CXX)
 #   - Set TRITON_PYTORCH_DOCKER_IMAGE to use the docker image of
 #     PyTorch to base the build off.
 #
+# TRITON_PYTORCH_ENABLE_TORCHTRT: enables Torch-TRT runtime support. Default is OFF.
+#   With TRITON_PYTORCH_DOCKER_IMAGE: should be set to ON for images 21.11+
+#   Without TRITON_PYTORCH_DOCKER_IMAGE:
+#     If you set it to ON: make sure to add Torch-TRT runtime lib path to TRITON_PYTORCH_LIB_PATHS. 
+#
+
```
CoderHam, line 49: Remove extra empty line

CMakeLists.txt
```diff
@@ -41,8 +41,16 @@ project(tritonpytorchbackend LANGUAGES C CXX)
...
 option(TRITON_ENABLE_GPU "Enable GPU support in backend" ON)
 option(TRITON_ENABLE_STATS "Include statistics collections in backend" ON)
+option(TRITON_PYTORCH_ENABLE_TORCHTRT "Enable TorchTRT support" OFF)
+
```
CoderHam, line 53: Remove extra empty line

CMakeLists.txt
```diff
@@ -279,7 +314,7 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
   )
 endif() # TRITON_PYTORCH_DOCKER_BUILD
 
-set(TRITON_PYTORCH_LDFLAGS "")
+set(TRITON_PYTORCH_LDFLAGS "-L.")
```
CoderHam, line 317: Why is this change needed?

CoderHam (changes requested):

CMakeLists.txt
CoderHam: Yes 21.12. Understood, we can set the default to ON once we have sufficient testing. I will make a note of the same.

borisfom: I have made it automatic (Cmake configured based on presence of Torch-TRT in PyTorch image).

CMakeLists.txt
borisfom: I am getting link error from libtorchtrt_runtime.so telling that this library is missing.

CMakeLists.txt
borisfom: Fixed manually.

CMakeLists.txt
borisfom: Must be a merge error. I will remove that.

CMakeLists.txt
```diff
@@ -199,6 +231,7 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
     PROPERTIES
       IMPORTED_LOCATION libtorch.so
   )
+
```
CoderHam, line 234: Remove extra empty line

CMakeLists.txt
```diff
@@ -149,6 +158,26 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
...
+  else()
+     set(TRITON_PYTORCH_ENABLE_TORCHTRT OFF)
+  endif()
+  
+  if (TRITON_PYTORCH_ENABLE_TORCHTRT)
+    set(PT_LIBS
+        ${PT_LIBS}
+        "libtorchtrt_runtime.so"
+    )
+  endif() #TRITON_PYTORCH_ENABLE_TORCHTRT
```
CoderHam, line 180: Space after `#`

CoderHam (changes requested):

CMakeLists.txt
borisfom: That would not work right if we set the default to ON, would it ?

CMakeLists.txt
borisfom: Right now OFF is a default setting. How do I tell default OFF from user-specified ?

borisfom:
Reworked the default: now it's ON, and user can set it to OFF and it will stay off. 
If set/left to ON, it would be disabled if the image does not have Torch-TRT, or if the local lib path does not contain installed Tocrh-TRT package location.

CMakeLists.txt
borisfom: Reworked the default: now it's ON, and user can set it to OFF and it will stay off.
If set/left to ON, it would be disabled if the image does not have Torch-TRT, or if the local lib path does not contain installed Tocrh-TRT package location.

CMakeLists.txt
borisfom: Re Jetson build - currently, we do not build pytorch_backend at all, so no changes are needed at this time.

CMakeLists.txt
CoderHam: Yes that should be good for now

CMakeLists.txt
CoderHam: But we should still have this check like we do not the ORT backend (as I pointed to before)

CMakeLists.txt
CoderHam: Why do you think it would not work right? The GPU is enabled by default as well. If the user disabled GPU support, we should require him to turn off Torch TRT support as well.

borisfom: I think it's a burden on the user - if he uses ON default, to require also specifically  add TRITON_PYTORCH_ENABLE_TORCHTRT=OFF - an option he obviously does not care about. And I can't tell if the user specified it to be on or have used the default.

borisfom: Look at the new logic - most of those diffs are outdated.

borisfom: Now the logic is : enabled by default, and we do checks if it's actually available, and turn off if it's not. Also, user can explicitly turn it OFF.

CMakeLists.txt
borisfom: No we should not  - now it's like 'off only' switch if user wants to turn it off. User never should specify it if he wants it ON.

CoderHam: How does an off only switch work? If the user specifies TRITON_ENABLE_GPU=OFF we want to throw an error and not enable TRITON_PYTORCH_ENABLE_TORCHTRT just because the pytorch container includes the torch trt libs.

CoderHam (changes requested):

CMakeLists.txt
```diff
@@ -150,6 +160,23 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
 
   string(REPLACE ";" " " CONDA_LIBS_STR "${CONDA_LIBS}")
 
+  if (${TRITON_PYTORCH_ENABLE_TORCHTRT})
+    execute_process(
```
CoderHam, line 164: @borisfom you should remove the auto detecting of TorchTRT and changing the cmake flag. We should **not silently change** the setting to make it valid. The pytorch backend should follow the pattern that other backends have and be consistent with how they handle that. Eg: https://github.com/triton-inference-server/onnxruntime_backend/blob/main/CMakeLists.txt
ORT backend as well has a flag for enabling TRT support: TRITON_ENABLE_ONNXRUNTIME_TENSORRT and has the checks I have asked you to add. It can be ON by default. But if that setting conflicts with other settings, like TRITON_ENABLE_GPU, then we should raise an error. 

Please make these changes so that we adhere to the existing patterns/policies Triton has.

CoderHam (changes requested):

CMakeLists.txt
borisfom: Not enable - Yes. Error-no. Unless I can tell user set TRITON_PYTORCH_ENABLE_TORCHTRT to ON explicitly, not just used all defaults ? Unless we can distinguish them, I think the best course of action is to adjust  the flags silently.

CMakeLists.txt
borisfom: "It can be ON by default. But if that setting conflicts with other settings, like TRITON_ENABLE_GPU, then we should raise an error." - that's a usability problem. It has to be OFF by default if you want to enforce it.

borisfom: I don't think TRITON_PYTORCH_ENABLE_TORCHTRT has to be consistent with TRITON_ENABLE_ONNXRUNTIME_TENSORRT - those are very different settings. For ONNXruntime - it changes basic execution, and therefore is OFF by default. TRITON_PYTORCH_ENABLE_TORCHTRT is only extending support, not changing anything. So we think it has to be ON by default. This does not even have to be an option - the only reason user may want to turn it OFF when it's available is to save 500K of memory if he is sure he is not going to need it.

borisfom: We can treat them the same though - if you think we should be consistent here, I will flip the default to OFF and we'll set the flag to ON in server's build.py. I would leave autodetect-off-switch on and issue a warning to handle the case someone would be building with latest code using PyTorch image from a few months back.

CMakeLists.txt
CoderHam: >  Unless I can tell user set TRITON_PYTORCH_ENABLE_TORCHTRT to ON explicitly, not just used all defaults ?

My perspective is the same that the user should control the option and we should not change these settings automatically. 
@deadeyegoodwin as well believes we should not adjust the flags silently.

CMakeLists.txt
CoderHam: I understand but based on the Triton policies/patterns we should be not setting these options automatically and that is a design choice the Triton team has taken and wishes to stick to.

CoderHam (changes requested):

borisfom:
I have left detection of not having Torch-TRT in and turning off the fag with a warning to address possible use of recent code with older images. Now that TRITON_PYTORCH_ENABLE_TORCHTRT is being set by build.py in server repo - how you suggest we address that? Will user have to edit it in server repo ?

CMakeLists.txt
borisfom: This is actually very good question - how exactly user set this flag when building the whole server?  Currently I am setting it in server's build.py but it's probably not the best idea - should be set in tritonserver repo.

CMakeLists.txt
```diff
@@ -150,6 +159,25 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
...
+      COMMAND /bin/sh -c "rm -fr /tmp/ttrt.so; docker rm -f pytorch_backend_ptlib > /dev/null 2>&1 || true; docker create --name pytorch_backend_ptlib ${TRITON_PYTORCH_DOCKER_IMAGE} && docker cp pytorch_backend_ptlib:/opt/conda/lib/python3.8/site-packages/torch_tensorrt/lib/libtorchtrt_runtime.so /tmp/ttrt.so && docker rm pytorch_backend_ptlib"
+      RESULT_VARIABLE TORCHTRT_STATUS
+      OUTPUT_VARIABLE TORCHTRT_OUT
+      ERROR_VARIABLE  TORCHTRT_ERR
+    )
+    # message("TORCHTRT_STATUS is ${TORCHTRT_STATUS}\nTORCHTRT_OUT is ${TORCHTRT_OUT}\nTORCHTRT_ERR is ${TORCHTRT_ERR}")
```
CoderHam, line 169: cleanup

CoderHam (changes requested):
@borisfom you will need to modify the build.py in server repo accordingly:
```
def pytorch_cmake_args(images):
    if "pytorch" in images:
        image = images["pytorch"]
    else:
        image = 'nvcr.io/nvidia/pytorch:{}-py3'.format(
            FLAGS.upstream_container_version)

    cargs = [
        '-DTRITON_PYTORCH_DOCKER_IMAGE={}'.format(image),
    ]
    if FLAGS.enable_gpu:
        cargs.append('-DTRITON_PYTORCH_ENABLE_TORCHTRT=ON')
    else:
        cargs.append('-DTRITON_ENABLE_GPU=OFF')

    return cargs
```

deadeyegoodwin:
@borisfom @CoderHam  There seems to be some disagreement in option handling. I haven't been following too closely but we have a general policy in triton that we need to follow. You can argue that Triton should have a different general policy, but at this point we're not going to change that so we just need to follow the policy we have.

An option can default to either on or off, whichever is more appropriate. The user can set the option to override that default. Regardless of how the option value is set, we either apply the option as specified, or we give an error as to why the option cannot be applied. For example, if GPU is disabled and someone tries to enable TensorRT we do not silently disable TensorRT, we don't disable TensorRT and print a warning... we raise an error and abort the build. The general concept is that we do exactly what the user requests and fail when doing so is not possible.

CMakeLists.txt
borisfom: @CoderHam : yes that's exactly what I did for server - that does not let the user override the option though, and would fail if an older Pytorch image is used (I do see pipelines running now for 21.08 etc).

CoderHam (approved):

PR #45, opened 2021-11-19
Return error if no inputs / outputs are provided in model config

CoderHam:
@deadeyegoodwin we now report an error if there are no inputs or outputs. However the ValidateInputs/ValidateOutputs functions do not check for the presence of inputs/outputs.

 I believe we should enforce this checks for IO validation.

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -225,29 +225,6 @@ ModelState::AutoCompleteConfig()
        Name() + "': not supported for pytorch backend")
           .c_str());
 
-  // If the model configuration doesn't specify any inputs or outputs
-  // then return error since pytorch backend does not support auto-completion.
-  size_t input_cnt = 0;
-  size_t output_cnt = 0;
-  {
-    triton::common::TritonJson::Value inputs;
-    if (ModelConfig().Find("input", &inputs)) {
-      input_cnt = inputs.ArraySize();
-    }
-
-    triton::common::TritonJson::Value outputs;
-    if (ModelConfig().Find("output", &outputs)) {
-      output_cnt = outputs.ArraySize();
-    }
-  }
-
-  if ((input_cnt == 0) || (output_cnt == 0)) {
-    return TRITONSERVER_ErrorNew(
-        TRITONSERVER_ERROR_INTERNAL,
-        "model configuration must contain inputs and outputs since "
-        "auto-complete is not supported for pytorch backend");
-  }
-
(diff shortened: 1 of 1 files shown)


src/libtorch.cc
```diff
@@ -225,6 +225,29 @@ ModelState::AutoCompleteConfig()
...
+      output_cnt = outputs.ArraySize();
+    }
+  }
+
+  if ((input_cnt == 0) || (output_cnt == 0)) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL,
+        "model configuration must contain inputs and outputs since "
+        "auto-complete is not supported for pytorch backend");
+  }
```
GuanLuo, line 249: The above should be remove as the validate functions are already responsible for checking if there are I/Os. Nothing should be done in autofill as PyTorch model doesn't provide model information.

src/libtorch.cc
```diff
@@ -501,7 +501,7 @@ ModelInstanceState::ModelInstanceState(
     }
   }
 
-  THROW_IF_BACKEND_INSTANCE_ERROR(ValidateInputs());
+  THROW_IF_BACKEND_INSTANCE_ERROR(ValidateInputs(expected_input_cnt));
```
GuanLuo, line 504: That won't be correct as `expected_input_cnt` looks at the model config already. So no input results in  `expected_input_cnt == 0` and validate function won't report error correctly

CoderHam: We have an additional check for when no inputs are specified: https://github.com/triton-inference-server/pytorch_backend/blob/77adac248889f21035eb57efcae5eae8d65ef5c0/src/libtorch.cc#L604-L608

Using expected_input_cnt here is symmetric with how the [onnxruntime backend](https://github.com/triton-inference-server/onnxruntime_backend/blob/main/src/onnxruntime.cc#L1134-L1141) does it.

CoderHam: Removed this check since we cannot use it for pytorch backend

GuanLuo: In ORT it is comparing the model config to the model metadata, which is not the same in PyTorch

GuanLuo:
Can you add this edge case to L0_model_config?

CoderHam:
> Can you add this edge case to L0_model_config?

https://github.com/triton-inference-server/server/pull/3600

GuanLuo (approved):

PR #49, opened 2022-01-24
Use tensor device instead of model device

CoderHam:
- Some models outputs can be on a device different than the model itself

Error before:
```
pinned buffer: failed to perform CUDA copy: invalid argument
```

Fixes: https://github.com/triton-inference-server/server/issues/2641

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -1165,6 +1165,9 @@ ModelInstanceState::ReadOutputTensors(
     const char* output_buffer =
         static_cast<const char*>(output_flat.data_ptr());
 
+    // Output tensors may not reside on the same device as model
+    torch::Device tensor_device = output_flat.device();
+
     //  Set output shape
     std::vector<int64_t> batchn_shape;
     auto shape = output_tensors[op_index].sizes();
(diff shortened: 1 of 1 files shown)


deadeyegoodwin (approved):
Is there a specific bug files that hit this case? Can we add a testcase for this?

CoderHam:
There is no bug for this. I noticed it just now while combing through the code and thought to fix it.

Yes we should add a test case for this with a custom model.

GuanLuo (approved):

tanmayv25 (approved):

CoderHam:
Hold on merge till test coverage is added by DLIS-3393

ddukek22:
👋  I was directed here from another issue. Thanks for all of the work on the project! Really appreciate it. Would it make sense to alert the user to the fact that this is happening transparently when the device of the output does not match the configured device for the model? It seems like there are scenarios where there could be performance implications that would be hard to track down if there wasn't really a trace of what was happening.

CoderHam:
@ddukek22 Triton will assume the device the tensor is returned on is the intended device and proceed as expected. It is the users responsibility to design the model with the right performance considerations. In some cases users want the device output to be on cpu. 

As for performance issues within Triton for such models:
Since Triton anyway copies the memory from GPU to CPU to create the response and return it via the network protocol the user used - http or grpc. The performance impact would be minimal.

ddukek22:
@CoderHam I agree with all of that. I was just stating that a debug scenario could prove difficult in a scenario where it wasn't intended for the tensor to end up on a device other than the one configured. Again thanks for all the work on the project!

PR #50, opened 2022-02-01
Add support for passing inputs as a dictionary to model

CoderHam:
- Only works for when the model has 1 input type of kind DictType.
- For such a model the input names need not use the `name__index` convention. 
- The names of the inputs are used as the key for the input dictionary.

Sample error message:

```
+----------------------------------+---------+-----------------------------------------------------------------+
| Model                            | Version | Status                                                          |
+----------------------------------+---------+-----------------------------------------------------------------+
| libtorch_list_dict               | 1       | UNAVAILABLE: Internal: An input of type 'Dict(str, Tensor)[]' w |
|                                  |         | as detected in the model. Only a single input of type Dict(str, |
|                                  |         |  Tensor) or input(s) of type Tensor are supported.              |
+----------------------------------+---------+-----------------------------------------------------------------+
```

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -1,4 +1,4 @@
-// Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -393,7 +393,7 @@ class ModelInstanceState : public BackendModelInstance {
   TRITONSERVER_Error* ValidateTypedSequenceControl(
       triton::common::TritonJson::Value& sequence_batching,
       const std::string& control_kind, bool required, bool* have_control);
-  TRITONSERVER_Error* ValidateInputs();
+  TRITONSERVER_Error* ValidateInputs(const size_t expected_input_cnt);
   TRITONSERVER_Error* ValidateOutputs();
   void Execute(
       std::vector<TRITONBACKEND_Response*>* responses,
@@ -506,7 +506,7 @@ ModelInstanceState::ModelInstanceState(
     }
   }
 
-  THROW_IF_BACKEND_INSTANCE_ERROR(ValidateInputs());
+  THROW_IF_BACKEND_INSTANCE_ERROR(ValidateInputs(expected_input_cnt));
   THROW_IF_BACKEND_INSTANCE_ERROR(ValidateOutputs());
 }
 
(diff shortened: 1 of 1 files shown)


src/libtorch.cc
```diff
@@ -589,6 +592,18 @@ ModelInstanceState::ValidateTypedSequenceControl(
 TRITONSERVER_Error*
 ModelInstanceState::ValidateInputs()
 {
+  const torch::jit::Method& method = torch_model_->get_method("forward");
+  const auto& schema = method.function().getSchema();
+  const std::vector<c10::Argument>& arguments = schema.arguments();
+
+  // Currently we only support Dict[str, Tensor] as input when there is a single
```
deadeyegoodwin, line 599: We support 1 dict argument, or any number of tensor arguments. You should check for both cases because we now know how to do it. That way if user does something other than the 2 cases that we support, you can give a clear error message.

deadeyegoodwin (changes requested):

src/libtorch.cc
CoderHam: Sure, will make that improvement.

CoderHam: Done. Also added a check for matching number of expected inputs between model and the config file.

src/libtorch.cc
```diff
@@ -430,6 +430,9 @@ class ModelInstanceState : public BackendModelInstance {
   // that output in the model.
   std::unordered_map<std::string, int> output_index_map_;
   std::unordered_map<std::string, TRITONSERVER_DataType> output_dtype_map_;
+
+  // If the input to the tensor is a dictionary of tensors.
```
tanmayv25, line 434: This comment is confusing. Shouldn't it be "If the input to the model is a dictionary of tensors"?

deadeyegoodwin (dismissed):

src/libtorch.cc
```diff
@@ -587,8 +590,49 @@ ModelInstanceState::ValidateTypedSequenceControl(
...
 
 TRITONSERVER_Error*
-ModelInstanceState::ValidateInputs()
+ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
 {
+  const torch::jit::Method& method = torch_model_->get_method("forward");
+  const auto& schema = method.function().getSchema();
+  const std::vector<c10::Argument>& arguments = schema.arguments();
+
+  // Currently we only support input of type Dict[str, Tensor] when the
```
tanmayv25, line 599: "Currently, we only support models with a single input of type Dict[str, Tensor]."
The way it is currently written means that pytorch backend assumes input of type Dict[str, Tesnor] if there is a single input in the model.

src/libtorch.cc
```diff
@@ -587,8 +590,49 @@ ModelInstanceState::ValidateTypedSequenceControl(
...
+  }
+  if ((arguments.size() == (1 + start_idx)) &&
+      (arguments.at(start_idx).type()->kind() == c10::TypeKind::DictType)) {
+    is_dict_input_ = true;
+  } else if (arguments.size() > start_idx) {
+    for (size_t i = start_idx; i < arguments.size(); i++) {
+      if (arguments.at(i).type()->kind() != c10::TypeKind::TensorType) {
+        return TRITONSERVER_ErrorNew(
+            TRITONSERVER_ERROR_INTERNAL,
+            "The model must expect a single input of type Dict[str, Tensor] "
```
tanmayv25, line 618: Error can be improved. "Multiple non-c10::TypeKind::TensorType inputs detected in the model. Only single  c10::TypeKind::DictType input is supported besides c10::TypeKind::TensorType inputs".  Or somehting like this

src/libtorch.cc
```diff
@@ -587,8 +590,49 @@ ModelInstanceState::ValidateTypedSequenceControl(
...
+  //
+  // Ignore the argument at idx 0 if it is of Class type (self param in forward
+  // function)
+  size_t start_idx = 0;
+  if ((arguments.size() > 0) &&
+      (arguments.at(0).type()->kind() == c10::TypeKind::ClassType)) {
+    start_idx = 1;
+  }
+  if ((arguments.size() == (1 + start_idx)) &&
+      (arguments.at(start_idx).type()->kind() == c10::TypeKind::DictType)) {
```
tanmayv25, line 611: Add a block for other c10::Typekinds and throw appropriate error.

tanmayv25 (changes requested):

GuanLuo (approved):

src/libtorch.cc
CoderHam: I want to use Dict[str, Tensor] since the Dict must have string and Tensor as key and value. The reason for not using c10::TypeKind is that I want this to match the convention used by PyTorch when logging the graph.

```
method forward {
  input self    ,__torch__.AddSubNet
  input input0.1        ,Dict(str, Tensor)
  }
 ```

src/libtorch.cc
CoderHam: The error is thrown in the section below where I check if the inputs are of type Tensor

CoderHam:
@tanmayv25 check out the sample error message when the input is a list of dict

src/libtorch.cc
tanmayv25: Ahh.. I see you have `arguments.size() > start_idx` instead of `arguments.size() > 1 + start_idx` below.

tanmayv25: Wouldn't it be better to return different errors for multiple Dict(str, Tensor) inputs vs unsupported TypeKind?

tanmayv25 (dismissed):

CoderHam:
@deadeyegoodwin @GuanLuo @tanmayv25 can one of you re-approve? I needed to rebase

deadeyegoodwin (approved):

tanmayv25 (approved):

PR #52, opened 2022-03-15
Disable NvFuser till fixed (Caused by move to 22.03 PyTorch)

CoderHam:


src/libtorch.cc
```diff
@@ -350,7 +351,8 @@ ModelState::ParseParameters()
         TRITONSERVER_ErrorDelete(err);
       }
     } else {
-      enable_nvfuser_pair_ = {true, enable_nvfuser};
+      // Override, disable NvFuser till fixed
+      enable_nvfuser_pair_ = {true, false /* enable_nvfuser */};
```
tanmayv25, line 355: Shouldn't we atleast emit a warning saying the nvfuser is being disabled instead of silently disabling it?

CoderHam: Corrected to update enable_nvfuser, that way the log in L357 will accurately reflect if the NvFuser has been disabled.

CoderHam: @tanmayv25 would you rather this be a Warning or an Info log?

tanmayv25: I would prefer an  additional warning. This way user would get to know why their preferred configuration were not selected.

src/libtorch.cc
```diff
@@ -335,6 +335,7 @@ ModelState::ParseParameters()
               .c_str());
     }
 
+    // TODO Re-enable NvFuser once fixed
```
GuanLuo, line 338: Make sure we have ticket to track re-enabling NvFuser

src/libtorch.cc
CoderHam: It probably doesn't make sense to have two logs so I am changing the current Log to warning for now.

src/libtorch.cc
CoderHam: I am using the same ticket. It is DLIS-3560.

src/libtorch.cc
```diff
@@ -350,9 +351,11 @@ ModelState::ParseParameters()
         TRITONSERVER_ErrorDelete(err);
       }
     } else {
+      // Override, disable NvFuser till fixed
+      enable_nvfuser = false;
```
Tabrizian, line 355: I think if the issue is in 22.03 pytorch and the default value for NVFuser is already false, we should not be forcing false here. Instead, we can add a release note for using NVFuser in Pytorch backend and mention that it doesn't work because of an issue in Pytorch.

CoderHam: The issue is that many users will update the Triton container version they use, then see a failure and open a bug. 

Only for us to then point them to the release notes. I rather will avoid any cases where a known error can cause issues in model loading / starting Triton server

Tabrizian: Ok, I'll be resolving this thread.

tanmayv25 (approved):

CoderHam:
TODO: ensure picked up by main when moving forward to using 22.03 upstream containers

PR #54, opened 2022-03-22
Share Model Weights for Instances on the Same Device

dyastremsky:
This PR allows a Triton user to enable multiple instances of a model on the same device to share weights. This is turned off by default and can be enabled via a model config parameter, "ENABLE_WEIGHT_SHARING."

Enabling weight reuse can reduce memory usage of model loading and inference. It should not be used with models that maintain state (due to the reusing of weights).

Related server test change: https://github.com/triton-inference-server/server/pull/4123

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -79,6 +79,10 @@ class ModelState : public BackendModel {
       std::string* model_path,
       std::shared_ptr<torch::jit::script::Module>* torch_model);
 
+  // Remove model from model map using its key
+  // Returns the number of models removed (0 or 1)
+  size_t UnloadModel(std::pair<bool, int64_t> key);
+
   bool EnabledOptimizedExecution() { return enable_optimized_execution_; }
   const std::pair<bool, bool>& EnabledTensorExprFuser() const
   {
@@ -123,8 +127,11 @@ class ModelState : public BackendModel {
   // Defaults to (false, false).
   std::pair<bool, bool> enable_nvfuser_pair_;
 
-  // Share torch module across instances on the same device. They key is a pair of isGPU, Device Index 
-  std::map<std::pair<bool, int64_t>, std::shared_ptr<torch::jit::script::Module>> torch_models_;
+  // Share torch module across instances on the same device. They key is a pair
+  // of isGPU, Device Index
+  std::map<
+      std::pair<bool, int64_t>, std::shared_ptr<torch::jit::script::Module>>
+      torch_models_;
 };
 
 TRITONSERVER_Error*
@@ -197,13 +204,17 @@ ModelState::LoadModel(
             "' for model instance '" + Name() + "'");
   }
 
-  // If torch model has already been been loaded on that device, skip loading that model 
-  // Else load the model as usual
+  // If torch model has already been been loaded on that device, skip loading
+  // that model Else load the model as usual
   auto device_pair = std::make_pair(!device.is_cpu(), device.index());
   auto mit = torch_models_.find(device_pair);
   if (mit != torch_models_.end()) {
     *torch_model = mit->second;
-    LOG_MESSAGE(TRITONSERVER_LOG_INFO, (std::string("Re-using model for pytorch for model instance '") + Name() + "'").c_str());
+    LOG_MESSAGE(
+        TRITONSERVER_LOG_INFO,
+        (std::string("Re-using model for pytorch for model instance '") +
+         Name() + "'")
+            .c_str());
     return nullptr;  // success
   }
 
@@ -230,6 +241,12 @@ ModelState::LoadModel(
   torch_models_.emplace(device_pair, *torch_model);
 
   return nullptr;  // success
+}
+
+size_t
+ModelState::UnloadModel(std::pair<bool, int64_t> key)
+{
+  return torch_models_.erase(key);
 }
 
 TRITONSERVER_Error*
@@ -528,8 +545,17 @@ ModelInstanceState::ModelInstanceState(
 
 ModelInstanceState::~ModelInstanceState()
 {
-  // Need to unload all instances on a device type before reset is called for the corr. torch_model_
-  // torch_model_.reset();
+  // If this is the last instance, remove the pointer from the map too to
+  // deallocate memory
+  if (torch_model_.use_count() == 2) {
+    if (StateForModel()->UnloadModel(
+            std::make_pair(!device_.is_cpu(), device_.index())) != 1) {
+      LOG_MESSAGE(
+          TRITONSERVER_LOG_ERROR,
+          ("Model mapping not found for '" + Name() + "'").c_str());
+    }
+  }
+  torch_model_.reset();
 #ifdef TRITON_ENABLE_GPU
   if (device_.is_cuda()) {
     c10::cuda::CUDACachingAllocator::emptyCache();
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -545,8 +545,7 @@ ModelInstanceState::ModelInstanceState(
 
 ModelInstanceState::~ModelInstanceState()
 {
-  // If this is the last instance, remove the pointer from the map too to
-  // deallocate memory
+  // If this is the last instance, remove the pointer from the model map
   if (torch_model_.use_count() == 2) {
     if (StateForModel()->UnloadModel(
             std::make_pair(!device_.is_cpu(), device_.index())) != 1) {
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -77,7 +77,7 @@ class ModelState : public BackendModel {
   TRITONSERVER_Error* LoadModel(
       const std::string& artifact_name, const torch::Device device,
       std::string* model_path,
-      std::unique_ptr<torch::jit::script::Module>* torch_model);
+      std::shared_ptr<torch::jit::script::Module>* torch_model);
 
   bool EnabledOptimizedExecution() { return enable_optimized_execution_; }
   const std::pair<bool, bool>& EnabledTensorExprFuser() const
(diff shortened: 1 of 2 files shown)


dyastremsky:
The code change is done. For the destructor, the memory is deallocated once the last shared pointer copy is removed. I'm going to verify the performance @CoderHam saw and confirm it is similar for stateful models.

I assume we don't need any new tests for this once performance is confirmed, since our CI will test for accuracy. Let me know if you want any new tests written.

CoderHam:
I'd recommend a dedicated test that confirms model weights are being shared / either by checking logs or by checking memory usage. 
@tanmayv25 does TRT have a similar test? TRT does share inference engines for multiple instances on the same device

tanmayv25:
In tensorrt backend, we do share a single engine on same device. We have tests for results accuracy in case of multiple instances but we don't have a dedicated test on just the sharing part of it.

@dyastremsky It will be good to have a test in this case to check whether we are sharing the model weights or not. Just checking the verbose logs should be okay. Also, make sure we have a test in our CI that verifies the results with multiple model instances and higher request concurrency.

CoderHam:
I agree that model weight sharing should be a configurable parameter and be disabled by default. I will do a more thorough review in a few hours.

src/libtorch.cc
```diff
@@ -512,6 +545,15 @@ ModelInstanceState::ModelInstanceState(
...
+  // If this is the last instance, remove the pointer from the model map
+  if (torch_model_.use_count() == 2) {
+    if (StateForModel()->UnloadModel(
+            std::make_pair(!device_.is_cpu(), device_.index())) != 1) {
+      LOG_MESSAGE(
+          TRITONSERVER_LOG_ERROR,
+          ("Model mapping not found for '" + Name() + "'").c_str());
+    }
+  }
   torch_model_.reset();
```
CoderHam, line 557: Does this not reset the shared pointer for all model instances on the same device?

CoderHam (changes requested):
Restructure to make model weight sharing optional. 

When model state is important, we should not share model weights so make this off by default

src/libtorch.cc
dyastremsky: No. It removes the current instance of shared pointer. When the last shared pointer is reset, the memory is deallocated.

I confirmed this via logging (using ::use_count to see the number of shared pointers) and C++ documentation: https://en.cppreference.com/w/cpp/memory/shared_ptr/reset
"If *this already owns an object and it is the last shared_ptr owning it, the object is destroyed through the owned deleter."

dyastremsky:
Thank you both for your feedback. It looks like we don't have tests with multiple instances and high concurrency, so I'll write one. I'll update the config reading to have a configurable "share_weights" parameter.

It doesn't look like logs report any weight-specific data, so I'll add some logs for model loading/unloading to check in the tests.

src/libtorch.cc
```diff
@@ -194,6 +204,20 @@ ModelState::LoadModel(
...
   }
 
+  // Skip loading model if it is already available on the target device
+  auto device_pair = std::make_pair(!device.is_cpu(), device.index());
+  auto mit = torch_models_.find(device_pair);
+  if (mit != torch_models_.end()) {
+    *torch_model = mit->second;
+    LOG_MESSAGE(
+        TRITONSERVER_LOG_INFO,
+        (std::string("Re-using model for pytorch for model instance '") +
```
CoderHam, line 214: Remove the "for pytorch" and reword to "Re-using TorchScript model for instance ..."

src/libtorch.cc
```diff
@@ -122,6 +126,12 @@ class ModelState : public BackendModel {
   // Flag pair to indicate whether nvfuser is set and enabled respectively.
   // Defaults to (false, false).
   std::pair<bool, bool> enable_nvfuser_pair_;
+
+  // Share torch module across instances on the same device. They key is a pair
```
CoderHam, line 130: Share TorchScript model across all instances on the same device. The key is a pair of isGPU and device index.

src/libtorch.cc
```diff
@@ -77,7 +77,11 @@ class ModelState : public BackendModel {
   TRITONSERVER_Error* LoadModel(
       const std::string& artifact_name, const torch::Device device,
       std::string* model_path,
-      std::unique_ptr<torch::jit::script::Module>* torch_model);
+      std::shared_ptr<torch::jit::script::Module>* torch_model);
+
+  // Remove model from model map using its key
+  // Returns the number of models removed (0 or 1)
+  size_t UnloadModel(std::pair<bool, int64_t> key);
```
CoderHam, line 84: If this will only return a 0 or 1 why not return a bool value instead? This bool would just mean if the model was unloaded or not.

src/libtorch.cc
```diff
@@ -512,6 +544,15 @@ ModelInstanceState::ModelInstanceState(
 
 ModelInstanceState::~ModelInstanceState()
 {
+  // If this is the last instance, remove the pointer from the model map
+  if (torch_model_.use_count() == 2) {
+    if (StateForModel()->UnloadModel(
+            std::make_pair(!device_.is_cpu(), device_.index())) != 1) {
+      LOG_MESSAGE(
+          TRITONSERVER_LOG_ERROR,
+          ("Model mapping not found for '" + Name() + "'").c_str());
```
CoderHam, line 553: "No model found on target device". I would suggest to log device type and device id here for readability.

Technically the mapping is not to a single model instance but rather a model shared across instances. So this statement is incorrect.

CoderHam (changes requested):

src/libtorch.cc
```diff
@@ -212,10 +246,26 @@ ModelState::LoadModel(
         TRITONSERVER_ERROR_INTERNAL,
         ("failed to load model '" + Name() + "': " + ex.what()).c_str());
   }
+  if (enable_weight_sharing_) {
```
CoderHam, line 249: Add empty line before this

CoderHam (changes requested):

CoderHam (dismissed):

dyastremsky:
@tanmayv25 @CoderHam Thank you both for your detailed comments. I addressed them and added tests in the associated server change. This PR is ready for re-review.

CoderHam (dismissed):
Nit: Update Readme copyright

CoderHam (dismissed):

src/libtorch.cc
```diff
@@ -512,6 +581,20 @@ ModelInstanceState::ModelInstanceState(
 
 ModelInstanceState::~ModelInstanceState()
 {
+  // If weight sharing is enabled and this is the last model instance,
+  // remove the pointer from the model map
+  if (torch_model_.use_count() == 2) {
```
GuanLuo, line 586: Why not just let `~ModelState()` to release all the models? The instances all share the same life cycle and when Triton unloads a model, the `TRITONBACKEND_Model` will be destroyed after all `TRITONBACKEND_ModelInstance` destruction.

I don't think you need to do anything explicitly as shared_ptr will handle that for you as long as the State objects are released.

dyastremsky: I'm not fully following. We have two options:
- Weights are not being shared: each model is pointed to by one instance
- Weights are being shared: each model is pointed to by each instance and is in the model mapping

It seems like the ModelState is shared across devices. We cannot store the model pointers in the ModelState, since there can be multiple devices (each with their own shared pointers). And weight sharing may be disabled, in which case each  model has its own pointer.

To delete the model, we need to release all copies of the shared pointer (in this case, one in each model and one in the model mapping; that changes if we move the model pointer to ModelState).

GuanLuo: I understand that there will be the mapping from device to torch model, what I am saying is that you don't need to explicitly perform `map.erase()` in `ModelState` and `shared_ptr.reset()` in `ModelInstanceState`. You need to delete the `torch_model` (managed in shared_ptr) only when Triton is unloading the model, so you can expect the `ModelInstanceState` and `ModelState` objects will be destroyed.

Consider the cases of sharing / not sharing with no explicit handling:

- Sharing: at least one `ModelInstanceState` holds shared pointer of the torch model; `ModelState` has another shared pointer in the map. On unloading model, `ModelInstanceState` destruction will destroy its copy of shared pointer, and then `ModelState` destruction will destroy the map which results in destroy the shared pointer in the map. Then all shared pointers of the torch model is destroyed and should automatically result in destroying the torch model.
- Not sharing: same as current behavior.

CoderHam: What is specific versions/instances of the model are being unloaded without the entire model being unloaded? For example if the user changes the instance_group field in the model config.pbtxt (with polling enabled)?

dyastremsky: I see, thank you for clarifying @GuanLuo.

For the edge cases @CoderHam pointed out:
- The instance_group case should work fine since the model does not change (each instance would have a pointer, the ModelState would have the same pointer).
- For versioning, each ModelState holds a different version, right? So if the model version is unloaded, the ModelState would be destroyed. If that's not the case, this PR would need to be updated to track version in the map (so model versions don't get shared).

I appreciate you thinking through the edge cases. If there's a scenario where the ModelState holds multiple models (e.g. versions) at the same time or the model gets updated without all ModelInstanceStates being destructed first, the code would need to be updated.

GuanLuo: Need to clarify that when it gets to the backend, "version" is no longer relevant. `TRITONBACKEND_Model` and the `TRITONBACKEND_ModelInstance` associated to it are just pointed to some files to load the model from, thus all instance will be "identical" since they are reading the model from the same set of files.

In other words, it is not the right scope to talk about "version" here, and all instances associated with a `TRITONBACKEND_Model` are simply be the copies of the same model. When instance group setting is updated (same for other file changes), think of it as creating new `TRITONBACKEND_Model` (and instances) and swap out the previous one.

dyastremsky: Done!

tanmayv25 (dismissed):

CoderHam (dismissed):

GuanLuo (approved):

dyastremsky:
I rebased. @GuanLuo @CoderHam @tanmayv25 would it be possible to get another review in real quick?

CoderHam (approved):

PR #56, opened 2022-03-31
Fix libc10 missing bug on jetson when loading pytorch backend

CoderHam:
Solves libc10.so not found, removes need to LD_LIBRARY_PATH to include pytorch backend dir and resolves segfault during shutdown when loading Onnx models.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -125,60 +125,67 @@ endif() # TRITON_ENABLE_GPU
...
 
-  string(REPLACE ";" " " CONDA_LIBS_STR "${CONDA_LIBS}")
+if (CMAKE_HOST_SYSTEM_PROCESSOR MATCHES "aarch64")
+  set(LIBS_ARCH "aarch64")
+  set(CONDA_LIBS
+      "libopenblas.so.0"
+  )
+else()
+  set(LIBS_ARCH "x86_64")
+  set(CONDA_LIBS
```
tanmayv25, line 158: CONDA_LIBS -> PT_CPU_LIBS

tanmayv25 (changes requested):

GuanLuo:
Do you need to have the same fix for main?

CMakeLists.txt
CoderHam: Technically this is for all dependencies that are inside the conda package/directory. Not just CPU libs. I would like to keep the name generic for this reason.

CoderHam:
> Do you need to have the same fix for main?

Yes. I will cherry-pick it over. As well as the change for Gitlab

tanmayv25 (approved):

PR #57, opened 2022-04-22
Remove __<index> from tensor name for dict input

baojunliu:
background:

For forward() with a dictionary, if we append __<index>  to input feature xyz, one will get feature xyz is not found in pytorch eval. If we removed __<index> in input spec of config.pbtxt and run python client without __<index>, the inference works.

However, when updating config.pbtxt for model with forward() using feature list, will get error during model loading.

When loading the model, we don't know if it is using dictionary or feature list. It seems the input handling is not consistent. 

This PR is to remove the __<index> from name.

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -1159,7 +1159,14 @@ ModelInstanceState::Execute(
       torch::Dict<std::string, torch::Tensor> input_dict;
       for (auto& input_index : input_index_map_) {
         torch::jit::IValue ival = (*input_tensors)[input_index.second];
-        input_dict.insert(input_index.first, ival.toTensor());
+        std::string tensor_name = input_index.first;
+        std::string deliminator = "__";
+        int start_pos = tensor_name.find(deliminator);
+        if (start_pos == -1) {
+          input_dict.insert(tensor_name, ival.toTensor());
+        } else {
+          input_dict.insert(tensor_name.substr(0, start_pos), ival.toTensor());
+        }
       }
       std::vector<torch::jit::IValue> input_dict_ivalue = {input_dict};
       model_outputs_ = torch_model_->forward(input_dict_ivalue);


CoderHam:
@baojunliu I am making some assumptions from the description here but it appears you are using a tool or some logic to generate the config input names. The fix this is making is to allow Triton to support the input names generated by the tool to be parsed into their corresponding dictionary key values (by removing the __index from the input name).

It appears that you are trying to add logic to triton to work around these cases.

The right way to do this would be to check that the model input type is a dictionary and change the input names for those cases from "<NAME__INDEX>" to just "<NAME>".

Feel free to correct my understanding. 

cc @dzier @msalehiNV

baojunliu:
@CoderHam

When loading the model, we don't know if it is using dictionary or feature list. If it is a dict, we need the "__<index>" to pass model loading, and remove it to make PyTorch eval work.

thuningxu:
@CoderHam isn't it a bit redundant to append __index to feature names, given that the index can be implied from `config.pbtxt` already? I think that can give you a much cleaner and unified config across different frameworks.

CoderHam:
> @baojunliu Reg: If it is a dict, we need the "__" to pass model loading, and remove it to make PyTorch eval work.

Is this limitation you are implying Triton has?

https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#special-conventions-for-pytorch-backend clarifies that the names for the inputs in the case of a single dictionary as input do not need `__<index>`. 


> @thuningxu Reg: isn't it a bit redundant to append __index to feature names, given that the index can be implied from config.pbtxt already? I think that can give you a much cleaner and unified config across different frameworks.

In the case where Triton enforced that the order for the input/output names must be the same i.e. `input__0` will always occur before `input__1` and so on, we would get rid of the index for the input/output names. 

The reason the configs cannot be identical across all frameworks is because the metadata PyTorch model stores/exposes is not very descriptive. However if you use the input names from the forward function as mentioned in the docs you would not need to use the __index convention for the inputs.

> If the forward function for the TorchScript model was defined as `forward(self, input0, input1)`, the first and second inputs could also be named "input0" and "input1" respectively.

baojunliu:
@CoderHam It requires different handling for dict and non-dict. We don't know if a model has dict. The workaround will let us to handle the input in a uniform way.


> > @baojunliu Reg: If it is a dict, we need the "__" to pass model loading, and remove it to make PyTorch eval work.
> 
> Is this limitation you are implying Triton has?
> 
> https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#special-conventions-for-pytorch-backend clarifies that the names for the inputs in the case of a single dictionary as input do not need `__<index>`.
>

thuningxu:
@CoderHam maybe I misunderstood something here. Do you mean `__<index>` is not required for `forward(self, input0, input1)` case? 

When is `__<index>` required for torchscript?

CoderHam:
> @CoderHam maybe I misunderstood something here. Do you mean __<index> is not required for forward(self, input0, input1) case?

Yes that is correct. Let me explain:

Let us assume a forward function that has the following definition:

```
forward(self, input0, input1)
```

The model config for the same (using only input names for simplicity):

```
input [
  {
    name: "input0"
    ...
  },
  {
    name: "input1"
    ...
  }
]
```

If you do not know the input names then in that case you can use the __index naming convention;

```
input [
  {
    name: "INPUT__0"
    ...
  },
  {
    name: "INPUT__1"
    ...
  }
]
```

Given this do you still feel there is a need for the enforcing of ordered inputs in the model config?

dyastremsky:
Closing PR due to inactivity. Feel free to reopen if further work is planned.

PR #60, opened 2022-04-27
Add support for String input/output

CoderHam:
- PyTorch does not natively support tensors of string
- We allow users to pass a list of type string instead. For now we allow only 1D lists of strings. This means that batching for string I/O is not supported.

TODO 
- [x] Add model generation and test coverage for addsub and identity models for string I/O

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,9 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
   configuration may help in some cases to avoid these negative interactions
   due to model specific caching and increase multiple instance performance.
 
-* Since PyTorch does not support Tensor of Strings, Triton allows users to pass
-string input(s)/recieve string output(s) using a List of Strings. Because of using
-List instead of Tensor, string I/O is supported only for 1-dimensional inputs/outputs.
+* PyTorch does not support Tensor of Strings but it does support models that accept
+a List of Strings as input(s) / produces a List of String as output(s). For these models
+Triton allows users to pass String input(s)/recieve String output(s) using the String
+datatype. As a limitation of using List instead of Tensor for String I/O, only for
+1-dimensional input(s)/output(s) are supported for I/O of String type.
 This means batching must be disabled as well for String I/O on TorchScript models.
--- a/README.md
+++ b/README.md
@@ -210,6 +210,6 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
   due to model specific caching and increase multiple instance performance.
 
 * Since PyTorch does not support Tensor of Strings, Triton allows users to pass
-string input/recieve string output using a List of Strings. Because of using List
-instead of Tensor, string I/O only is supported only for 1 dimension. This means
-that the `shape` must a single dimension and batching must be disabled as well.
+string input(s)/recieve string output(s) using a List of Strings. Because of using
+List instead of Tensor, string I/O is supported only for 1-dimensional inputs/outputs.
+This means batching must be disabled as well for String I/O on TorchScript models.
--- a/README.md
+++ b/README.md
@@ -214,4 +214,5 @@ a List of Strings as input(s) / produces a List of String as output(s). For thes
 Triton allows users to pass String input(s)/recieve String output(s) using the String
 datatype. As a limitation of using List instead of Tensor for String I/O, only for
 1-dimensional input(s)/output(s) are supported for I/O of String type.
-This means batching must be disabled as well for String I/O on TorchScript models.
+Batching is not allowed for PyTorch models with String I/O. For these models,
+the user must specify `max_batch_size: 0` in the configuration.
--- a/README.md
+++ b/README.md
@@ -208,3 +208,8 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
   Setting the parameter `DISABLE_OPTIMIZED_EXECUTION` to "true" in the model
   configuration may help in some cases to avoid these negative interactions
   due to model specific caching and increase multiple instance performance.
+
+* Since PyTorch does not support Tensor of Strings, Triton allows users to pass
+string input/recieve string output using a List of Strings. Because of using List
+instead of Tensor, string I/O only is supported only for 1 dimension. This means
+that the `shape` must a single dimension and batching must be disabled as well.
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -1197,6 +1197,197 @@ ModelInstanceState::Execute(
   }
 }
 
+// This function will return a tensor's contents as a contiguous
+// chunk in system memory. In some cases this will require copying the data.
+// If that  happens, 'contiguous_buffer' will be set to hold the contiguous
+// chunk and 'cuda_copy' will be set to indicate whether CUDA copy is
+// conducted.  The data copy can be avoided if the input is already in
+// a contiguous chunk and the input is located in memory type and id
+// specified.
+TRITONSERVER_Error*
+GetContiguousInputContent(
+    TRITONBACKEND_Input* rinput, const uint32_t buffer_count,
+    const char** content, size_t* content_byte_size, char** contiguous_buffer,
+    cudaStream_t stream, bool* cuda_copy)
+{
+  *cuda_copy = false;
+  *contiguous_buffer = nullptr;
+
+  // Check input buffers to see if data copy is necessary
+  size_t chunk_count = 0;
+  bool type_mismatch = false;
+  uint64_t total_byte_size = 0;
+  for (size_t idx = 0; idx < buffer_count; ++idx) {
+    TRITONSERVER_MemoryType src_memory_type;
+    int64_t src_memory_type_id;
+    size_t src_byte_size;
+    const void* src_ptr;
+
+    RETURN_IF_ERROR(TRITONBACKEND_InputBuffer(
+        rinput, idx, &src_ptr, &src_byte_size, &src_memory_type,
+        &src_memory_type_id));
+
+    if (src_ptr != nullptr) {
+      chunk_count++;
+      total_byte_size += src_byte_size;
+      type_mismatch |= (src_memory_type == TRITONSERVER_MEMORY_GPU);
+    }
+  }
+
+  if (chunk_count == 0) {
+    *content = nullptr;
+    *content_byte_size = 0;
+  } else if ((chunk_count == 1) && !type_mismatch) {
+    TRITONSERVER_MemoryType src_memory_type;
+    int64_t src_memory_type_id;
+    RETURN_IF_ERROR(TRITONBACKEND_InputBuffer(
+        rinput, 0, (const void**)content, content_byte_size, &src_memory_type,
+        &src_memory_type_id));
+  } else {
+    *contiguous_buffer = (char*)malloc(total_byte_size);
+
+    size_t offset = 0;
+    for (size_t i = 0; i < chunk_count; i++) {
+      bool cuda_used;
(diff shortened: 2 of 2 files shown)


CoderHam:
Tested locally and running CI. Working on TODOs

tanmayv25:
Should add a comment about the limitation of TYPE_STRING support here:
https://github.com/triton-inference-server/pytorch_backend#important-note

CoderHam:
> Should add a comment about the limitation of TYPE_STRING support here: https://github.com/triton-inference-server/pytorch_backend#important-note

Done

README.md
```diff
@@ -208,3 +208,8 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
   Setting the parameter `DISABLE_OPTIMIZED_EXECUTION` to "true" in the model
   configuration may help in some cases to avoid these negative interactions
   due to model specific caching and increase multiple instance performance.
+
+* Since PyTorch does not support Tensor of Strings, Triton allows users to pass
+string input/recieve string output using a List of Strings. Because of using List
```
tanmayv25, line 213: string input(s)/receive string output(s)?

tanmayv25: We can support multiple 1-D input tensors right?

README.md
```diff
@@ -208,3 +208,8 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
...
+instead of Tensor, string I/O only is supported only for 1 dimension. This means
```
tanmayv25, line 214: string I/O only is supported only for 1-dimensional inputs.

README.md
```diff
@@ -208,3 +208,8 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
...
+that the `shape` must a single dimension and batching must be disabled as well.
```
tanmayv25, line 215: This means batching must be disabled as well for String I/O on TorchScript models.

src/libtorch.cc
```diff
@@ -1284,62 +1569,113 @@ ModelInstanceState::ReadOutputTensors(
...
+    } else if (output_tensors[op_index].isList()) {
+        // Custom handling for string/bytes tensor...
+
+      torch::List<torch::jit::IValue> output_list = output_tensors[op_index].toList();
+
+      // Get output shape
+      size_t list_length = output_list.size();
+      std::vector<int64_t> batchn_shape{list_length};
+      if (output_list.elementType()->kind() != c10::TypeKind::StringType) {
+        return TRITONSERVER_ErrorNew(
```
tanmayv25, line 1641: In future would we want to support the list of integers/floating points?

src/libtorch.cc
```diff
@@ -1284,62 +1569,113 @@ ModelInstanceState::ReadOutputTensors(
...
+
+        tensor_offset += tensor_element_cnt;
+      }
+    } else {
+      return TRITONSERVER_ErrorNew(
           TRITONSERVER_ERROR_INVALID_ARG,
           (std::string("output '") + name +
-           "' is a scalar which is not supported.")
-              .c_str()));
+           "' must be of type Tensor or List[str].")
```
tanmayv25, line 1675: We must throw an error when loading the model.
1. If the dynamic_batching is enabled for string input.
2. If more than one dimension is specified for TYPE_STRING.

tanmayv25 (changes requested):

src/libtorch.cc
CoderHam: If we a separate process for supporting List of datatypes that are supported inside Torch. We would want to use the Complex Tensor Structure design that I had suggested earlier

README.md
CoderHam: Only twice?
Also do we want to mention inputs and outputs again after already mentioning String I/O?

README.md
CoderHam: Yes we can

src/libtorch.cc
```diff
@@ -1284,62 +1621,114 @@ ModelInstanceState::ReadOutputTensors(
...
+      }
 
-    if (batchn_shape.size() == 0) {
-      RETURN_IF_ERROR(TRITONSERVER_ErrorNew(
+      size_t tensor_offset = 0;
+
+      for (size_t idx = 0; idx < responses->size(); idx++) {
+        auto& response = (*responses)[idx];
+
+        const size_t tensor_element_cnt = GetElementCount(batchn_shape);
```
GuanLuo, line 1706: Shouldn't use element count for the whole batch, again, batching can be enable even for 1-D tensor, you can have a batch of [2], [1] and the output shape should be set accordingly instead of 3.

src/libtorch.cc
```diff
@@ -1194,6 +1246,249 @@ ModelInstanceState::Execute(
...
 
+// This function will return a tensor's contents as a contiguous
+// chunk in system memory. In some cases this will require copying the data.
+// If that  happens, 'contiguous_buffer' will be set to hold the contiguous
+// chunk and 'cuda_copy' will be set to indicate whether CUDA copy is
+// conducted.  The data copy can be avoided if the input is already in
+// a contiguous chunk and the input is located in memory type and id
+// specified.
+TRITONSERVER_Error*
+GetContiguousInputContent(
```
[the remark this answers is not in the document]
CoderHam: No we do not. We should add it in a different ticket. I believe this is useful for Tensorflow and ORT as well.

src/libtorch.cc
```diff
@@ -1170,12 +1223,11 @@ ModelInstanceState::Execute(
...
       auto model_outputs_tuple = model_outputs_.toTuple();
       for (auto& m_op : model_outputs_tuple->elements()) {
-        output_tensors->push_back(m_op.toTensor());
+        output_tensors->push_back(m_op);
       }
     } else {
       try {
-        auto model_output_tensor = model_outputs_.toTensor();
-        output_tensors->push_back(model_output_tensor);
+        output_tensors->push_back(model_outputs_);
```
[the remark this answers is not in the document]
CoderHam: Fixed to raise appropriate exception.

src/libtorch.cc
```diff
@@ -1194,6 +1246,249 @@ ModelInstanceState::Execute(
...
+                          "expected " + std::to_string(request_element_cnt) +
+                          " strings for inference input '" + name + "', got " +
+                          std::to_string(element_idx))
+                          .c_str()));
+    FillStringTensor(
+        input_list, tensor_offset + element_idx,
+        request_element_cnt - element_idx);
+  }
+
+  free(contiguous_buffer);
```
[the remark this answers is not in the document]
CoderHam: Followed the approach we use for tensorflow backend. The memory must only be release conditionally when there is an error.

src/libtorch.cc
CoderHam: The batchn_shape is derived from the actual output of the pytorch model. It would always be the right shape.

src/libtorch.cc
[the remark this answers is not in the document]
GuanLuo: I may have missed that but `free()` is called before every return, under what condition the memory won't be released?

src/libtorch.cc
GuanLuo: See [comment](https://github.com/triton-inference-server/pytorch_backend/pull/60/files#r863965574) that the model can allow batching even the input is 1-D string, then in my above example (assume identity), the model returns [3] for `batchn_shape`, I don't think here properly scatter the output to different responses.

src/libtorch.cc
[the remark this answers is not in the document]
CoderHam: I had a second look and you are right. I have wrapped it in a unique_ptr. Thanks for catching this

src/libtorch.cc
tanmayv25: The actual scatter happens here: https://github.com/triton-inference-server/tensorflow_backend/blob/main/src/tensorflow.cc#L2059
We are anyways setting a check against dynamic batching while loading the model. I think we must not concern too much right now with DB if the user doesn't want it. We can add the feature later on?

CoderHam: I agree. Since we are close to code freeze and these changes are stable as is. I will file a ticket for dynamic batching and the reshaped + batching use case that @GuanLuo pointed out.

README.md
```diff
@@ -208,3 +208,8 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
   Setting the parameter `DISABLE_OPTIMIZED_EXECUTION` to "true" in the model
   configuration may help in some cases to avoid these negative interactions
   due to model specific caching and increase multiple instance performance.
+
+* Since PyTorch does not support Tensor of Strings, Triton allows users to pass
+string input(s)/recieve string output(s) using a List of Strings. Because of using
+List instead of Tensor, string I/O is supported only for 1-dimensional inputs/outputs.
+This means batching must be disabled as well for String I/O on TorchScript models.
```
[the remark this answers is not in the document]
CoderHam: I want to do this and the required test coverage for this in a follow up PR since I would like to get these changes in before code freeze. Reshaping for string lists will be different than the way we do it for tensors. 
Related: https://github.com/triton-inference-server/pytorch_backend/pull/60#discussion_r866184080

CoderHam:
@tanmayv25 @GuanLuo for re-reviews

src/libtorch.cc
```diff
@@ -819,13 +848,38 @@ ModelInstanceState::ValidateOutputs()
...
+      // validating the model shapes.
+      std::vector<int64_t> dims;
+      triton::common::TritonJson::Value reshape;
+      if (io.Find("reshape", &reshape)) {
+        RETURN_IF_ERROR(ParseShape(reshape, "shape", &dims));
+      } else {
+        RETURN_IF_ERROR(ParseShape(io, "dims", &dims));
+      }
+
+      if ((dims.size() + (supports_batching ? 1 : 0)) > 1) {
```
GuanLuo, line 872: Since this can't support batching at all, should be `if ((dims.size() > 1)  || supports_batching)`, otherwise the reshape case will pass the validation

src/libtorch.cc
```diff
@@ -1194,6 +1265,246 @@ ModelInstanceState::Execute(
...
+  size_t element_idx = 0;
+
+  // For string data type, we always need to have the data on CPU so
+  // that we can read string length and construct the string
+  // properly. So if the request's input tensor is not in CPU need to
+  // copy it there.
+  const char* content = nullptr;
+  size_t content_byte_size = 0;
+
+  std::unique_ptr<char> contiguous_buffer;
```
GuanLuo, line 1371: Doesn't unique_ptr call `delete`? The pointer is allocated via `malloc` and should be released via `free`. https://stackoverflow.com/questions/10854210/behaviour-of-malloc-with-delete-in-c

Why not just use `std::vector<char>` and store the contiguous buffer? You don't need to use smart pointer since the lifecycle is not going beyond this function.

README.md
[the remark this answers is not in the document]
GuanLuo: Should say something like "batching is not allowed for TorchScript models with String I/O, `max_batch_size: 0` must be set for such models". It is more about we haven't decided on how to represent batching with respect to Python string.

src/libtorch.cc
CoderHam: Done

src/libtorch.cc
CoderHam: Done

README.md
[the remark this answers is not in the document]
CoderHam: Done

GuanLuo (approved):

tanmayv25 (approved):

PR #61, opened 2022-04-30
Add CUDA cache cleaning flag to pytorch backend

zeruniverse:
Implemented a flag similar to `memory_arena_shrinkage` in ONNXRuntime backend. With `ENABLE_CACHE_CLEANING` set to `true`, CUDA cache gets cleared after each model execution. This will enable us to serve more models in one GPU. 

I have tested on local GPU: with `enable_cache_cleaning=false`, the model takes additional 2Gi GRAM after running inference with batch size 30. After that, this 2Gi never gets shrunk. With `enable_cache_cleaning=true`, the additional GRAM gets shrunk every time after model execution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,9 @@ key: "ENABLE_WEIGHT_SHARING"
 
 * `ENABLE_CACHE_CLEANING`: Boolean flag to enable CUDA cache cleaning after each model execution.
 If not specified, cache cleaning is disabled. This flag has no effect if model is on CPU.
+Setting this flag to true will negatively impact the performance due to additional CUDA cache
+cleaning operation after each model execution. Therefore, you should only use this flag if you
+serve multiple models with Triton and encounter CUDA out of memory issue during model executions.
 
 The section of model config file specifying this parameter will look like:
 
--- a/README.md
+++ b/README.md
@@ -179,6 +179,20 @@ key: "ENABLE_WEIGHT_SHARING"
 }
 ```
 
+* `ENABLE_CACHE_CLEANING`: Boolean flag to enable CUDA cache cleaning after each model execution.
+If not specified, cache cleaning is disabled. This flag has no effect if model is on CPU.
+
+The section of model config file specifying this parameter will look like:
+
+```
+parameters: {
+key: "ENABLE_CACHE_CLEANING"
+    value: {
+    string_value:"true"
+    }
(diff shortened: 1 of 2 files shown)


manhtd98:
why we not merge it? i have the same problem

tanmayv25:
This might have fell through the cracks. Thanks for bringing it up.

src/libtorch.cc
```diff
@@ -445,6 +469,9 @@ class ModelInstanceState : public BackendModelInstance {
   void ProcessRequests(
       TRITONBACKEND_Request** requests, const uint32_t request_count);
 
+  // Clear cache
```
tanmayv25, line 472: Clear CUDA cache

src/libtorch.cc
```diff
@@ -1551,6 +1583,8 @@ TRITONBACKEND_ModelInstanceExecute(
   // specific request.
   instance_state->ProcessRequests(requests, request_count);
 
+  if(model_state->EnabledCacheCleaning()) instance_state->ClearCache();
```
tanmayv25, line 1586: enclose the call  within braces.

tanmayv25 (changes requested):

README.md
```diff
@@ -179,6 +179,20 @@ key: "ENABLE_WEIGHT_SHARING"
 }
 ```
 
+* `ENABLE_CACHE_CLEANING`: Boolean flag to enable CUDA cache cleaning after each model execution.
+If not specified, cache cleaning is disabled. This flag has no effect if model is on CPU.
```
tanmayv25, line 183: You should also talk about how it would penalize the performance.

zeruniverse:
@tanmayv25 do you have a CLA version for individual contributor? This CLA is for corporation.

tanmayv25:
@zeruniverse  Just provide your name instance of company name in the form. We use the same form for company and individual.

zeruniverse:
@tanmayv25 

- [x] CLA sent to triton-cla at nvidia dot com

- [x] addressed review comments

- [ ] perf data (do we need it? I have no hardware to do a thorough perf test)

tanmayv25:
Thanks @zeruniverse! I will add some sanity testing for this on our CI and merge the changes.

tanmayv25 (approved):

PR #63, opened 2022-05-17
Enforce ordering of I/O if naming convention is not followed

CoderHam:
- Add fall back to enforcing ordering of inputs/outputs for cases when neither naming convention for I/O is met i.e. not an input argument in the model forward function and not using the <name>__<index> format for names.
- These models that failed to follow the naming conventions would earlier fail to load. Now they load successfully and the input/outputs assume the order they are specified in inside the model configuration file.

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -424,6 +424,14 @@ ModelState::ParseParameters()
   return nullptr;
 }
 
+// The naming convention followed for inputs/outputs in the model configuration.
+// Outputs don't support ARGUEMENT.
+enum NamingConvention {
+  NAMED_INDEX,
+  FORWARD_ARGUEMENT,
+  STRICT_CONFIG_ORDERING
+};
+
 //
 // ModelInstanceState
 //
@@ -475,6 +483,11 @@ class ModelInstanceState : public BackendModelInstance {
       TRITONBACKEND_Request** requests, const uint32_t request_count,
       std::vector<TRITONBACKEND_Response*>* responses,
       uint64_t* compute_end_ns);
+
+  // Validate naming convention for inputs and outputs in config
+  TRITONSERVER_Error* GetNamingConvention(
+      NamingConvention* naming_convention,
+      const std::set<std::string>& allowed_io);
 
   ModelState* model_state_;
 
@@ -651,10 +664,6 @@ ModelInstanceState::ValidateTypedSequenceControl(
   return nullptr;  // success
 }
 
-// The naming convention followed for inputs/outputs in the model configuration.
-// Outputs don't support ARGUEMENT.
-typedef enum { INDEX, ARGUEMENT, STRICT_ORDERING } NamingConvention;
-
 TRITONSERVER_Error*
 ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
 {
@@ -732,58 +741,8 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
 
   bool supports_batching = model_state_->MaxBatchSize() > 0;
 
-  NamingConvention naming_convention = NamingConvention::ARGUEMENT;
-
-  // Rules for (non-Dictionary) input tensor names:
-  // 1. Must be in 'allowed_inputs' (arguments in the forward function)
-  // 2. Must follow the naming convention i.e. <name>__<index>
-  // 3. If neither of the above conditions are satisfied, enforce strict
-  // ordering of model inputs.
-  for (size_t i = 0; i < ios.ArraySize(); i++) {
-    triton::common::TritonJson::Value io;
-    RETURN_IF_ERROR(ios.IndexAsObject(i, &io));
-
-    // Validate name
-    std::string io_name;
-    RETURN_IF_ERROR(io.MemberAsString("name", &io_name));
-    auto itr = allowed_inputs.find(io_name);
-    if (itr == allowed_inputs.end()) {
-      naming_convention = NamingConvention::STRICT_ORDERING;
-      break;
-    }
-  }
-
-  // If not, check if inputs follow INDEX
-  if (naming_convention == NamingConvention::STRICT_ORDERING) {
-    for (size_t i = 0; i < ios.ArraySize(); i++) {
-      triton::common::TritonJson::Value io;
-      RETURN_IF_ERROR(ios.IndexAsObject(i, &io));
-
-      // Validate name
-      std::string io_name;
-      RETURN_IF_ERROR(io.MemberAsString("name", &io_name));
-      try {
-        int start_pos = io_name.find(deliminator);
-        if (start_pos == -1) {
-          throw std::invalid_argument("input must follow naming convention");
-        }
-        ip_index = std::atoi(io_name.substr(start_pos + 2).c_str());
-        naming_convention = NamingConvention::INDEX;
-      }
-      catch (std::exception& ex) {
-        LOG_MESSAGE(
-            TRITONSERVER_LOG_WARN,
-            ("input '" + io_name + "' is neither an input argument to the " +
-             "model '" + model_state_->Name() +
-             "' nor does it follow the naming convention i.e. "
-             "<name>__<index>. Falling back to enforcing ordering from model "
-             "configuration.")
-                .c_str());
-        naming_convention = NamingConvention::STRICT_ORDERING;
-        break;
-      }
-    }
-  }
+  NamingConvention naming_convention;
+  RETURN_IF_ERROR(GetNamingConvention(&naming_convention, allowed_inputs));
 
   for (size_t i = 0; i < ios.ArraySize(); i++) {
     triton::common::TritonJson::Value io;
@@ -796,27 +755,37 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
       // If dictionary, index is irrelevant but we use the map to store the
       // input names since they are the keys for the dictionary
       input_index_map_[io_name] = i;
-    } else if (naming_convention == NamingConvention::ARGUEMENT) {
-      auto itr = allowed_inputs.find(io_name);
-      if (itr != allowed_inputs.end()) {
-        input_index_map_[io_name] = std::distance(allowed_inputs.begin(), itr);
-      }
-    } else if (naming_convention == NamingConvention::INDEX) {
-      try {
-        int start_pos = io_name.find(deliminator);
-        if (start_pos == -1) {
-          throw std::invalid_argument("input must follow naming convention");
-        }
-        ip_index = std::atoi(io_name.substr(start_pos + 2).c_str());
-        input_index_map_[io_name] = ip_index;
-      }
-      catch (std::exception& ex) {
-        // Already ran through this logic earlier. Try block will not catch an
-        // exception.
-      }
     } else {
-      // 'naming_convention' is STRICT_ORDERING
-      input_index_map_[io_name] = i;
+      switch (naming_convention) {
+        case FORWARD_ARGUEMENT: {
+          auto itr = allowed_inputs.find(io_name);
+          if (itr != allowed_inputs.end()) {
+            input_index_map_[io_name] =
+                std::distance(allowed_inputs.begin(), itr);
+          }
+          break;
+        }
+        case NAMED_INDEX: {
+          try {
+            int start_pos = io_name.find(deliminator);
+            if (start_pos == -1) {
+              throw std::invalid_argument(
+                  "input must follow naming convention");
+            }
+            ip_index = std::atoi(io_name.substr(start_pos + 2).c_str());
+            input_index_map_[io_name] = ip_index;
+          }
+          catch (std::exception& ex) {
+            // Already ran through this logic earlier. Try block will not catch
+            // an exception.
+          }
+          break;
+        }
+        case STRICT_CONFIG_ORDERING: {
+          input_index_map_[io_name] = i;
+          break;
+        }
+      }
     }
 
     // Validate data type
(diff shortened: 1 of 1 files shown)


src/libtorch.cc
```diff
@@ -740,8 +740,11 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
       // input names since they are the keys for the dictionary
       input_index_map_[io_name] = i;
     } else {
-      // input tensor name must be in 'allowed_inputs' or must follow the naming
-      // convention
+      // Rules for (non-Dictionary) input tensor names:
+      // 1. Must be in 'allowed_inputs' (arguments in the forward function)
+      // 2. Must follow the naming convention i.e. <name>__<index>
+      // 3. If neither of the above conditions are satisfied, enforce strict
+      // ordering of model inputs.
```
GuanLuo, line 747: Shouldn't you check and decide which of the rules should be applied for all inputs (and outputs below) first? What happens if user provides a config with 3. in mind but the name happens to match one of the `allowed_inputs`

src/libtorch.cc
```diff
@@ -740,8 +740,11 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
       // input names since they are the keys for the dictionary
       input_index_map_[io_name] = i;
     } else {
-      // input tensor name must be in 'allowed_inputs' or must follow the naming
-      // convention
+      // Rules for (non-Dictionary) input tensor names:
+      // 1. Must be in 'allowed_inputs' (arguments in the forward function)
+      // 2. Must follow the naming convention i.e. <name>__<index>
+      // 3. If neither of the above conditions are satisfied, enforce strict
```
dyastremsky, line 746: Should this say "if both of the above conditions aren't satisfied"? Or some variation. Right now, it sounds like strict ordering will be followed only if 1 and 2 are both not true.

dyastremsky (changes requested):

src/libtorch.cc
```diff
@@ -755,12 +758,15 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
...
-               "follow the naming convention i.e. <name>__<index>.")
+          LOG_MESSAGE(
+              TRITONSERVER_LOG_WARN,
+              ("input '" + io_name + "' is neither an input argument to the " +
+               "model '" + model_state_->Name() +
+               "' nor does it follow the naming convention i.e. "
+               "<name>__<index>. Falling back to enforcing ordering from model "
+               "configuration.")
                   .c_str());
+          input_index_map_[io_name] = i;
```
dyastremsky, line 769: Clarifying question: We're allowing users to potentially set a convention for some inputs/outputs but not others? I'm not sure if this is an intended side effect of this ticket, but wanted to confirm.

src/libtorch.cc
tanmayv25: I agree with fixing the edge case here. 
@CoderHam But can you remind me the reason behind this change? As in why  3 is needed if user can enforce ordering using the naming convention?

src/libtorch.cc
CoderHam: > Right now, it sounds like strict ordering will be followed only if 1 and 2 are both not true.
It is a fall back if neither of the two conditions are met.

src/libtorch.cc
CoderHam: > Shouldn't you check and decide which of the rules should be applied for all inputs (and outputs below) first? What happens if user provides a config with 3. in mind but the name happens to match one of the allowed_inputs

A user may want to use a mixture of the naming convention i.e. use input names from the model's forward function but random output names from the output. 

I agree that we can enforce that all the inputs (and outputs) follow the same naming convention. That being said we do not check for a collision. For example if the user specifies INPUTA__0 and INPUTB__0 the model would be loaded correctly but see issues during execution. Are you suggesting we add tests for such collisions?

CoderHam: > But can you remind me the reason behind this change? As in why 3 is needed if user can enforce ordering using the naming convention?

The reason for adding the third case is for users who want to specify arbitrary names for I/O. They need only know the number of Inputs and Outputs and not their names.

src/libtorch.cc
CoderHam: I was aware of this side effect and added testing for it as well. If the consensus of the team is to force a specific convention we can follow that. That being said we do not strictly validate IO names. In fact one can load a model with 2 inputs both as INPUT__0 at the moment.

src/libtorch.cc
tanmayv25: I think the ask here is how do you plan to handle cases where the model forward function has the following inputs in order:
1. INPUT0, INPUT1, INPUT2, INPUT3
But the input in config are following :

2. INPUT4, INPUT5, INPUT0, INPUT1

CoderHam: That would work without my new changes for cases where the inputs are already in the forward function, the order is extracted from the allowed_inputs i.e. the order in which they are present in the model's forward function.

src/libtorch.cc
dyastremsky: Perfect, thanks for explaining and testing.

dyastremsky (approved):

src/libtorch.cc
tanmayv25: INPUT4 and INPUT5 are not in the forward function right?

May be the mapping is(using 3): INPUT0->INPUT4, INPUT1->INPUT5, INPUT2->INPUT0, INPUT3->INPUT1
Or may be the mapping is(using partially 1): INPUT0->INPUT0, INPUT1-INPUT1, INPUT2->INPUT4, INPUT3->INPUT5

CoderHam: @tanmayv25 I have enforced the need for consistency of input / output naming conventions to avoid this issue. I have not added logic to detect collisions yet. For inputs there can be multi-way collisions due to the 3 choices for naming format.  

please re-review

src/libtorch.cc
```diff
@@ -651,6 +651,10 @@ ModelInstanceState::ValidateTypedSequenceControl(
   return nullptr;  // success
 }
 
+// The naming convention followed for inputs/outputs in the model configuration.
+// Outputs don't support ARGUEMENT.
+typedef enum { INDEX, ARGUEMENT, STRICT_ORDERING } NamingConvention;
```
tanmayv25, line 656: arguement -> forward_argument

tanmayv25, line 656: NamingConvention -> IOTensorOrder

INDEX, FORWARD, CONFIG??? 

It makes sens as we look at the name of the tensors to identify the order in which we must place the arguments.

tanmayv25 (changes requested):

src/libtorch.cc
GuanLuo: Use C++ style enum

src/libtorch.cc
```diff
@@ -728,6 +732,59 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
...
+             "model '" + model_state_->Name() +
+             "' nor does it follow the naming convention i.e. "
+             "<name>__<index>. Falling back to enforcing ordering from model "
+             "configuration.")
+                .c_str());
+        naming_convention = NamingConvention::STRICT_ORDERING;
+        break;
+      }
+    }
+  }
```
GuanLuo, line 786: Put above to a helper function i.e. `GetNamingConvention()` and it should be able to be re-used by output validation with some twists, i.e. passing `allow_inputs` as argument and use empty set for output to skip the check as there is no named output support

src/libtorch.cc
```diff
@@ -739,30 +796,27 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
       // If dictionary, index is irrelevant but we use the map to store the
       // input names since they are the keys for the dictionary
       input_index_map_[io_name] = i;
-    } else {
-      // input tensor name must be in 'allowed_inputs' or must follow the naming
-      // convention
+    } else if (naming_convention == NamingConvention::ARGUEMENT) {
```
GuanLuo, line 799: use `switch` since you have enum

GuanLuo (changes requested):

src/libtorch.cc
```diff
@@ -740,27 +756,34 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
...
+            if (start_pos == -1) {
+              throw std::invalid_argument(
+                  "input must follow naming convention");
+            }
+            ip_index = std::atoi(io_name.substr(start_pos + 2).c_str());
+            input_index_map_[io_name] = ip_index;
+          }
+          catch (std::exception& ex) {
+            // Already ran through this logic earlier. Try block will not catch
+            // an exception.
```
GuanLuo, line 780: Just remove the check as you know the I/Os are "good" after `GetNamingConvention()`

src/libtorch.cc
```diff
@@ -424,6 +424,14 @@ ModelState::ParseParameters()
   return nullptr;
 }
 
+// The naming convention followed for inputs/outputs in the model configuration.
+// Outputs don't support ARGUEMENT.
+enum NamingConvention {
```
GuanLuo, line 429: use `enum class` to be safer

src/libtorch.cc
```diff
@@ -1265,6 +1299,94 @@ ModelInstanceState::Execute(
...
+      // Validate name
+      std::string io_name;
+      RETURN_IF_ERROR(io.MemberAsString("name", &io_name));
+      try {
+        int start_pos = io_name.find(deliminator);
+        if (start_pos == -1) {
+          throw std::invalid_argument(
+              "input/output must follow naming convention");
+        }
+        io_index = std::atoi(io_name.substr(start_pos + 2).c_str());
```
GuanLuo, line 1360: Remove unused variable

GuanLuo: Okay, I missed the intention here. so you don't really need the result of the conversion so just remove `io_index` and call `std::atoi(io_name.substr(start_pos + 2).c_str());` without assigning the return value to anything. And can you add a comment on what it is actually doing here.

src/libtorch.cc
```diff
@@ -1265,6 +1281,93 @@ ModelInstanceState::Execute(
...
+               "Falling back to enforcing ordering from model configuration.")
+                  .c_str());
+        }
+        *naming_convention = NamingConvention::STRICT_CONFIG_ORDERING;
+        break;
+      }
+    }
+  }
+
+  return nullptr;  // success
```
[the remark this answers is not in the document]
CoderHam: The typo was introduced because my latest stashed changes were pushed. They are present now.

dyastremsky (dismissed):

GuanLuo (dismissed):

src/libtorch.cc
```diff
@@ -424,6 +424,14 @@ ModelState::ParseParameters()
   return nullptr;
 }
 
+// The naming convention followed for inputs/outputs in the model configuration.
+// Outputs don't support ARGUEMENT.
```
tanmayv25, line 428: ARGUEMENT-> ARGUMENT

src/libtorch.cc
```diff
@@ -424,6 +424,14 @@ ModelState::ParseParameters()
...
+enum class NamingConvention {
```
tanmayv25, line 429: So we are sticking with NamingConvention instead of IOTensorOrder?

src/libtorch.cc
tanmayv25: Here and everywhere. There are several places where you  have ARGUEMENT

src/libtorch.cc
CoderHam: Yes, in my opinion it is earlier to pair this with the Naming Convention detailed in the model configuration.

src/libtorch.cc
```diff
@@ -1265,6 +1284,96 @@ ModelInstanceState::Execute(
...
+      std::string io_name;
+      RETURN_IF_ERROR(io.MemberAsString("name", &io_name));
+      try {
+        int start_pos = io_name.find(deliminator);
+        if (start_pos == -1) {
+          throw std::invalid_argument(
+              "input/output must follow naming convention");
+        }
+        // Use atoi to check if the index part of the name is an integer
+        std::atoi(io_name.substr(start_pos + 2).c_str());
```
nv-kmcgill53, line 1347: I don't think I understand how this line checks the index. You call std::atoi but don't use the returned value. From [the documentation](https://en.cppreference.com/w/cpp/string/byte/atoi) it doesn't appear to throw an exception either.

src/libtorch.cc
```diff
@@ -1265,6 +1284,96 @@ ModelInstanceState::Execute(
...
+      RETURN_IF_ERROR(ios.IndexAsObject(i, &io));
+
+      // Validate name
+      std::string io_name;
+      RETURN_IF_ERROR(io.MemberAsString("name", &io_name));
+      try {
+        int start_pos = io_name.find(deliminator);
+        if (start_pos == -1) {
+          throw std::invalid_argument(
+              "input/output must follow naming convention");
```
nv-kmcgill53, line 1344: NIT: try-catching in c++ is pretty expensive even if this is as straight forward as it looks. I tried to make a [quick example](https://godbolt.org/z/hfTGceErK) of what the output of the assembler looks like. You can see that gcc has some more bloat with the exception handling as compared to the standard if-statement error handling. Turning on -O3 didn't really improve things either. If we expect multiple types of exceptions to be thrown in this block then we can keep it but otherwise I would suggest using the standard if-statements to handle this particular case.

src/libtorch.cc
```diff
@@ -829,19 +846,21 @@ ModelInstanceState::ValidateOutputs()
...
     RETURN_IF_ERROR(io.MemberAsString("name", &io_name));
-    try {
-      int start_pos = io_name.find(deliminator);
-      if (start_pos == -1) {
-        throw std::invalid_argument("output must follow naming convention");
+    switch (naming_convention) {
+      case NamingConvention::NAMED_INDEX: {
+        int start_pos = io_name.find(deliminator);
+        if (start_pos == -1) {
+          throw std::invalid_argument("output must follow naming convention");
```
nv-kmcgill53, line 853: Didn't this get validated already in GetNamingConvention() line 1340?

CoderHam: Yes good catch

src/libtorch.cc
CoderHam: I think is a great idea. I have re-worked to avoid the catching of exceptions for std::aoti

src/libtorch.cc
CoderHam: Not needed anymore since atoi is not needed here

CoderHam: PS: It appears testing this that the exception is indeed caught even if no value is returned

nv-kmcgill53: This appears to be part of the undefined behavior then, or is undocumented: [No-throw guarantee: this function never throws exceptions.](https://www.cplusplus.com/reference/cstdlib/atoi/). Either way, not a good idea when compiling for different systems :)

nv-kmcgill53 (approved):

tanmayv25 (approved):

PR #64, opened 2022-05-23
Add nvtx markers which can be useful in perf profiling

tanmayv25:
Added these markers for perf profiling for a bug. I think it is good idea to merge it in main for future to avoid repeated work.
See here for header: https://github.com/triton-inference-server/common/blob/main/include/triton/common/nvtx.h
Note: The GPU forward execution is an asynchronous operation in pytorch backend.

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -33,6 +33,7 @@
 #include "triton/backend/backend_model.h"
 #include "triton/backend/backend_model_instance.h"
 #include "triton/backend/backend_output_responder.h"
+#include "triton/common/nvtx.h"
 #include "triton/core/tritonbackend.h"
 
 #ifdef TRITON_PYTORCH_ENABLE_TORCHVISION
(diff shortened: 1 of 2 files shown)


GuanLuo (approved):

dyastremsky (approved):

src/libtorch.cc
```diff
@@ -897,6 +897,8 @@ ModelInstanceState::ProcessRequests(
        std::to_string(request_count) + " requests")
           .c_str());
 
+  NVTX_RANGE(nvtx_, "ProcessRequests " + Name());
```
rmccorm4, line 900: will these compile if TRITON_ENABLE_NVTX not defined?

rmccorm4: Ah, nevermind I see it will: https://github.com/triton-inference-server/common/blob/1603d152f2b090194183abe09034f4b9f62fbd49/include/triton/common/nvtx.h#L57

rmccorm4 (approved):

PR #66, opened 2022-06-23
Fix intel mkl issue that causes segfault

CoderHam:
- during destruction
- causes freeze during inference

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,6 +185,9 @@ set(OPENCV_LIBS
     "libjpeg.so"
 )
 
+# The patchelf commands ensure the MKL libraries are loaded correctly during runtime
+# Without these, the framework/backend complains of missing libraries / symbols and 
+# in some cases leads to segmentation faults.
 if (${TRITON_PYTORCH_DOCKER_BUILD})
   string(REPLACE ";" " " CONDA_LIBS_STR "${CONDA_LIBS}")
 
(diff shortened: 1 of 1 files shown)


mc-nv (dismissed):

CMakeLists.txt
```diff
@@ -228,6 +228,10 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
     COMMAND /bin/sh -c "if [ -f libmkl_avx2.so ]; then patchelf --add-needed libmkl_core.so libmkl_avx2.so; fi"
     COMMAND /bin/sh -c "if [ -f libmkl_avx512.so ]; then patchelf --add-needed libmkl_gnu_thread.so libmkl_avx512.so; fi"
     COMMAND /bin/sh -c "if [ -f libmkl_avx512.so ]; then patchelf --add-needed libmkl_core.so libmkl_avx512.so; fi"
+    COMMAND /bin/sh -c "if [ -f libmkl_vml_def.so ]; then patchelf --add-needed libmkl_gnu_thread.so libmkl_vml_def.so; fi"
```
tanmayv25, line 231: Add a comment describing why we are patchelf'ing these shared libraries here for the future reference.

tanmayv25 (changes requested):

CMakeLists.txt
tanmayv25: And how to obtain the dependency graph.

CoderHam: Done

tanmayv25 (approved):

CMakeLists.txt
CoderHam: @tanmayv25 obtaining the dependency graph is complicated and a matter of reviewing the mkl library and its inner workings. 

I have skipped adding the logic here since it requires using ldd, LD_DEBUG, nm and understanding how mkl works.

PR #69, opened 2022-06-30
Add reshape+batching and dynamic batching support for string I/O

krishung5:
Test added here: https://github.com/triton-inference-server/server/pull/4574

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,5 +231,3 @@ a List of Strings as input(s) / produces a List of String as output(s). For thes
 Triton allows users to pass String input(s)/recieve String output(s) using the String
 datatype. As a limitation of using List instead of Tensor for String I/O, only for
 1-dimensional input(s)/output(s) are supported for I/O of String type.
-Batching is not allowed for PyTorch models with String I/O. For these models,
-the user must specify `max_batch_size: 0` in the configuration.
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -1560,7 +1560,6 @@ SetStringInputTensor(
                   std::to_string(element_idx + 1) + " for inference input '" +
                   name + "', expecting " + std::to_string(request_element_cnt))
                   .c_str()));
-      FillStringTensor(input_list, request_element_cnt - element_idx);
       return cuda_copy;
     }
 
@@ -1600,7 +1599,9 @@ SetStringInputTensor(
                           " strings for inference input '" + name + "', got " +
                           std::to_string(element_idx))
                           .c_str()));
-    FillStringTensor(input_list, request_element_cnt - element_idx);
+    if (element_idx < request_element_cnt) {
+      FillStringTensor(input_list, request_element_cnt - element_idx);
+    }
   }
 
   return cuda_copy;
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -837,8 +837,7 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
               .c_str());
     }
 
-    // Validate shape for String inputs. Only allow 1 dimension and no
-    // batching.
+    // Validate shape for String inputs. Only allow 1 dimension.
     if (io_dtype == "TYPE_STRING") {
       // If a reshape is provided for the input then use that when
       // validating the model shapes.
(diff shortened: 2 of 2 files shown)


src/libtorch.cc
```diff
@@ -1532,9 +1527,7 @@ SetStringInputTensor(
                   std::to_string(element_idx + 1) + " for inference input '" +
                   name + "', expecting " + std::to_string(request_element_cnt))
                   .c_str()));
-      FillStringTensor(
-          input_list, tensor_offset + element_idx,
```
GuanLuo, line 1536: `tensor_offset` is no longer needed? If it becomes unused parameter, then you can remove it from the function signature and propagate the change up on the call stack.

src/libtorch.cc
```diff
@@ -1832,6 +1813,7 @@ ModelInstanceState::ReadOutputTensors(
 
     } else if (output_tensors[op_index].isList()) {
       // Custom handling for string/bytes tensor...
+      const int max_batch_size = model_state_->MaxBatchSize();
```
GuanLuo, line 1816: You can store `support_batching` as member variable, then you don't need to compute the value again and again

README.md
```diff
@@ -214,5 +214,3 @@ a List of Strings as input(s) / produces a List of String as output(s). For thes
 Triton allows users to pass String input(s)/recieve String output(s) using the String
 datatype. As a limitation of using List instead of Tensor for String I/O, only for
 1-dimensional input(s)/output(s) are supported for I/O of String type.
-Batching is not allowed for PyTorch models with String I/O. For these models,
```
CoderHam, line 217: We still cannot allow 2D string I/O in the pytorch framework. Modify the docs to highlight the same

src/libtorch.cc
CoderHam: +1

CoderHam (changes requested):

README.md
krishung5: I think this should be addressed already in the previous line where we mention the framework only supports 1D string I/O?

src/libtorch.cc
krishung5: Thanks for the catch. Removed the unused variable.

src/libtorch.cc
krishung5: Done

src/libtorch.cc
```diff
@@ -1532,9 +1528,7 @@ SetStringInputTensor(
                   std::to_string(element_idx + 1) + " for inference input '" +
                   name + "', expecting " + std::to_string(request_element_cnt))
                   .c_str()));
-      FillStringTensor(
-          input_list, tensor_offset + element_idx,
-          request_element_cnt - element_idx);
+      FillStringTensor(input_list, request_element_cnt - element_idx);
```
GuanLuo, line 1531: I don't think you need to fill the rest of string elements in this case, when `element_idx >= request_element_cnt` evaluates to true, there is `element_idx+1` elements in `input_list`

src/libtorch.cc
```diff
@@ -1576,9 +1568,7 @@ SetStringInputTensor(
                           " strings for inference input '" + name + "', got " +
                           std::to_string(element_idx))
                           .c_str()));
-    FillStringTensor(
-        input_list, tensor_offset + element_idx,
-        request_element_cnt - element_idx);
+    FillStringTensor(input_list, request_element_cnt - element_idx);
```
GuanLuo, line 1571: Same here, only need to fill if `element_idx < request_element_cnt`, so add an additional check right before `FillStringTensor`

src/libtorch.cc
krishung5: Thanks for the catch. Removed this line.

src/libtorch.cc
krishung5: Added a check for this

krishung5:
CI passes: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/5274652 please review, thanks!

CoderHam (approved):

GuanLuo (approved):

PR #73, opened 2022-07-22
Fix multiple instance performance

Tabrizian:


GuanLuo:
FYI, the title of the graph are swapped.

src/libtorch.cc
```diff
@@ -552,11 +582,48 @@ ModelInstanceState::Create(
...
 ModelInstanceState::ModelInstanceState(
     ModelState* model_state, TRITONBACKEND_ModelInstance* triton_model_instance)
+#ifdef TRITON_ENABLE_GPU
+    : BackendModelInstance(model_state, triton_model_instance),
+      model_state_(model_state), device_(torch::kCPU), is_dict_input_(false),
+      pytorch_stream_(at::cuda::getDefaultCUDAStream())
+#else
     : BackendModelInstance(model_state, triton_model_instance),
       model_state_(model_state), device_(torch::kCPU), is_dict_input_(false)
+#endif
```
GuanLuo, line 592: Change to below? Or initialize `pytorch_stream_` in constructor body

```suggestion
    : BackendModelInstance(model_state, triton_model_instance),
      model_state_(model_state), device_(torch::kCPU), is_dict_input_(false)
#ifdef TRITON_ENABLE_GPU
      , pytorch_stream_(at::cuda::getDefaultCUDAStream())
#endif
```

Tabrizian:
@GuanLuo good catch! Fixed.

src/libtorch.cc
```diff
@@ -552,11 +582,48 @@ ModelInstanceState::Create(
...
 {
   if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
     device_ = torch::Device(torch::kCUDA, DeviceId());
```
GuanLuo, line 595: Should this be also in the ENABLE_GPU condition?

src/libtorch.cc
```diff
@@ -552,11 +582,48 @@ ModelInstanceState::Create(
...
+#ifdef TRITON_ENABLE_GPU
+    pytorch_stream_ =
+        at::cuda::getStreamFromPool(false /* is_high_priority */, DeviceId());
```
GuanLuo, line 598: Does this guarantee to not returning stream that being used by other instances? I don't see places to release the stream back to the pool.

And what is the difference between using this stream v.s. the one we created in backend utils (`stream_`), would it achieve the same if we wrap `stream_` into `at::cuda::CUDAStream` if the library exposes API to do so?

src/libtorch.cc
```diff
@@ -552,11 +582,48 @@ ModelInstanceState::Create(
...
+    if (stream_ != nullptr) {
+      cudaError_t err = cudaStreamDestroy(stream_);
+      if (err != cudaSuccess) {
+        TRITONSERVER_LogMessage(
+            TRITONSERVER_LOG_ERROR, __FILE__, __LINE__,
+            (std::string("~BackendModelInstance: ") + name_ +
+             " failed to destroy cuda stream: " + cudaGetErrorString(err))
+                .c_str());
+      }
+      stream_ = pytorch_stream_.stream();
```
GuanLuo, line 608: Should be careful here, `~BackendModelInstance` will destroy the stream if `stream_ != nullptr`, would that interference with Torch's internal management?

src/libtorch.cc
```diff
@@ -1056,6 +1125,15 @@ ModelInstanceState::ProcessRequests(
   std::vector<BackendMemory*> input_memories;
   bool cuda_copy = false;
   std::unique_ptr<BackendInputCollector> collector;
+  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+#ifdef TRITON_ENABLE_GPU
+    RESPOND_ALL_AND_SET_TRUE_IF_ERROR(
+        responses, request_count, all_response_failed,
+        ConvertCUDAStatusToTritonError(
+            cudaEventRecord(compute_input_event_, stream_),
```
GuanLuo, line 1133: Shouldn't the event be recorded after all input copies are issued?

src/libtorch.cc
```diff
@@ -1070,52 +1148,41 @@ ModelInstanceState::ProcessRequests(
...
-          break;
-        }
-
-        output_names.emplace_back(io_name);
-      }
+  // If the instance kind is not GPU, we need to synchronize the CUDA stream
+  if (Kind() != TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+#ifdef TRITON_ENABLE_GPU
+    if (cuda_copy) {
+      cudaStreamSynchronize(stream_);
```
GuanLuo, line 1155: Can sync by event

src/libtorch.cc
```diff
@@ -552,11 +582,48 @@ ModelInstanceState::Create(
 
 ModelInstanceState::ModelInstanceState(
     ModelState* model_state, TRITONBACKEND_ModelInstance* triton_model_instance)
+#ifdef TRITON_ENABLE_GPU
+    : BackendModelInstance(model_state, triton_model_instance),
+      model_state_(model_state), device_(torch::kCPU), is_dict_input_(false),
+      pytorch_stream_(at::cuda::getDefaultCUDAStream())
```
tanmayv25, line 588: Why is 'pytorch_stream_' initialization is needed here? What happens if `pytorch_stream_` remains uninitialized for CPU instances?

tanmayv25: If important, can we not set `pytorch_stream_=at::cuda::getDefaultCUDAStream()` within the function below? It will avoid a lot of code duplication.

src/libtorch.cc
```diff
@@ -1155,19 +1222,75 @@ ModelInstanceState::ProcessRequests(
   }
 
   uint64_t compute_end_ns = 0;
+  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+#ifdef TRITON_ENABLE_GPU
```
GuanLuo, line 1226: Seems redundant to have it everywhere, can this be encapsulated somehow
```
  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
#ifdef TRITON_ENABLE_GPU
```

src/libtorch.cc
tanmayv25, line 598: From the documentation it looks like this call only returns cuda stream from a pre-allocated pool. Where are the cuda streams in the pool being created?
https://pytorch.org/cppdocs/api/function_namespacec10_1_1cuda_1a42eb10631a7944924acd387aeae13245.html

I think it will make more sense to use high_priority cuda stream for inference if the stream is already been created in the pool. Have you run your experiments with is_high_priority=true?
I wonder if we can expose that as an option in model config.

src/libtorch.cc
Tabrizian: Yes, I have fixed this in the next commit.

tanmayv25, line 608: I think we must not set  stream_ here like this. Triton would assume the ownership of stream here.

src/libtorch.cc
```diff
@@ -1070,52 +1151,41 @@ ModelInstanceState::ProcessRequests(
...
-        const char* io_name;
-        size_t io_name_len;
-        err = io.MemberAsString("name", &io_name, &io_name_len);
-        if (err != nullptr) {
-          break;
-        }
-
-        output_names.emplace_back(io_name);
-      }
+  // If the instance kind is not GPU, we need to synchronize the CUDA stream
```
tanmayv25, line 1154: This is counter-intuitive. For a non-GPU instance, why do we need to synchronize the CUDA stream?

src/libtorch.cc
nv-kmcgill53: The high priority option looks like it gets a stream from another pool, is this high priority pool a part of the same pre-allocation? Are these high priority streams local to the model instance or global to the device?

src/libtorch.cc
GuanLuo: The request data can be from GPU and we copy to CPU, need explicit sync as the model execution will not be on the stream used on copy

Tabrizian: Exactly.

src/libtorch.cc
Tabrizian: The problem is that `pytorch_stream_` is not default constructible. Because of this it must be initialized in the initializer list. I also tried using a unique_ptr or a pointer to the return argument of `at::cuda::getDefaultCUDAStream()`. However, it is an r-value and you cannot get the pointer to an r-value. Let me know if you have any other ideas.

src/libtorch.cc
Tabrizian: I will ask the pytorch team to make sure but I think as long as the object is not destructed it will not be returned to the pool. I think the release will happen when the object is destructed.

> And what is the difference between using this stream v.s. the one we created in backend utils (stream_), would it achieve the same if we wrap stream_ into at::cuda::CUDAStream if the library exposes API to do so?

I think at::cuda::CUDAStream is just a lightweight wrapper around the cudaStream. There is now way to create a PyTorch cudaStream from our own cudaStream. That is why I have used PyTorch's stream as our stream for input/output collectors.

src/libtorch.cc
Tabrizian: This mainly marks the beginning of the input collection. The input compute part is the duration between `compute_input_event_` and `compute_infer_event_`.

src/libtorch.cc
Tabrizian: This is for the case where the PyTorch stream is not created (i.e model is using KIND_CPU) so the events do not exist.

src/libtorch.cc
Tabrizian: See my response here: https://github.com/triton-inference-server/pytorch_backend/pull/73#discussion_r930512176

src/libtorch.cc
Tabrizian: The ownership is handled correctly. We set the stream_ to null before the instance is destructed.

src/libtorch.cc
Tabrizian: I have fixed this in the next commit.

src/libtorch.cc
Tabrizian: Fixed.

src/libtorch.cc
Tabrizian: Switched to getstreamfromexternal https://pytorch.org/cppdocs/api/function_namespacec10_1_1cuda_1a399b2870dd4f6e4f9424666762080332.html#function-c10-cuda-getstreamfromexternal.

This API is more recently added. We construct a PyTorch stream using the model instance stream.

src/libtorch.cc
Tabrizian: I have fixed the duplication issue.

src/libtorch.cc
Tabrizian: Switched to model instance CUDA instances instead of using CUDA . I have not run any experiment using the is_high_priority=true.

> Where are the cuda streams in the pool being created?

It is created in PyTorch: https://github.com/pytorch/pytorch/blob/f9ef363982136f45dfb2bd4205c545cb17e59afd/c10/cuda/CUDAStream.cpp#L23-L50
Looks like there are 32 streams per GPU per pool.

> Have you run your experiments with is_high_priority=true?

I have not run the experiments with high priority=true. It looks like TRT also uses CUDAStreams with priorities? I was wondering whether you have seen any speedup gains from using streams in TRT. https://github.com/triton-inference-server/tensorrt_backend/blob/main/src/tensorrt.cc#L3064-L3077

I can file a follow up ticket to investigate this and expose this option in backend utilities or model configuration.

src/libtorch.cc
Tabrizian: Refactored.

src/libtorch.cc
GuanLuo: after we can use `stream_` for execution with the FromExternal, event is only used to keep track of the duration of different compute phase?

Tabrizian: Correct.

src/libtorch.cc
```diff
@@ -1155,19 +1192,81 @@ ModelInstanceState::ProcessRequests(
...
+    RESPOND_ALL_AND_SET_TRUE_IF_ERROR(
+        responses, request_count, all_response_failed,
+        ConvertCUDAStatusToTritonError(
+            cudaEventElapsedTime(
+                &compute_output_duration, compute_output_event_,
+                compute_output_end_event_),
+            TRITONSERVER_ERROR_INTERNAL, "Failed to capture elapsed time"));
+    compute_start_ns = exec_start_ns + (compute_input_duration * 1e6);
+    compute_end_ns = compute_start_ns + (compute_infer_duration * 1e6);
+    exec_end_ns = compute_end_ns + (compute_output_duration * 1e6);
```
GuanLuo, line 1265: I think `exec_end_ns` can just be the timestamp captured at this place and no CUDA event is needed. Then we can move the calculation after the responses are sent

src/libtorch.cc
```diff
@@ -1070,56 +1126,37 @@ ModelInstanceState::ProcessRequests(
...
+          &compute_start_ns,
+          reinterpret_cast<cudaEvent_t*>(&compute_infer_event_)));
 
   // Run...
   if (!all_response_failed) {
-    Execute(&responses, request_count, &input_tensors, &output_tensors);
+    if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+#ifdef TRITON_ENABLE_GPU
+      at::cuda::CUDAStream stream =
+          at::cuda::getStreamFromExternal(stream_, DeviceId());
```
GuanLuo, line 1153: Can we keep this as member variable so we don't construct Torch stream wrapper inflight.

src/libtorch.cc
```diff
@@ -530,6 +553,14 @@ class ModelInstanceState : public BackendModelInstance {
 
   // If the model supports batching.
   bool supports_batching_;
+
+#ifdef TRITON_ENABLE_GPU
+  // PyTorch stream used for execution of inferences.
+  cudaEvent_t compute_input_event_;
```
tanmayv25, line 559: compute_input_start_event_

src/libtorch.cc
```diff
@@ -530,6 +553,14 @@ class ModelInstanceState : public BackendModelInstance {
...
+  cudaEvent_t compute_infer_event_;
```
tanmayv25, line 560: compute_infer_start_event_

src/libtorch.cc
```diff
@@ -530,6 +553,14 @@ class ModelInstanceState : public BackendModelInstance {
...
+  cudaEvent_t compute_output_event_;
```
tanmayv25, line 561: compute_output_start_event_

src/libtorch.cc
```diff
@@ -1070,56 +1126,37 @@ ModelInstanceState::ProcessRequests(
...
+          reinterpret_cast<cudaEvent_t*>(&compute_infer_event_)));
 
   // Run...
   if (!all_response_failed) {
-    Execute(&responses, request_count, &input_tensors, &output_tensors);
+    if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+#ifdef TRITON_ENABLE_GPU
+      at::cuda::CUDAStream stream =
+          at::cuda::getStreamFromExternal(stream_, DeviceId());
+      at::cuda::CUDAStreamGuard device_guard{stream};
```
tanmayv25, line 1154: With the same line of reasoning, can we not declare CUDAStreamGuard once for the instance object so that we don't have to construct it with every instance? 
Also, we would not have to declare the stream guard again in L1205-L1207. We would need not replicate the #ifdef conditions at two place. 
Is there are reason for constructing a new CudaStreamGuard right before these calls?

src/libtorch.cc
Tabrizian: I agree. I have fixed this.

src/libtorch.cc
Tabrizian: Putting it in the model instance object didn't work but I used another API which is thread local and is working properly.

GuanLuo (approved):

tanmayv25 (dismissed):

Tabrizian:
Pipeline passed: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/5554052

Tabrizian:
Test passed: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/5576070

GuanLuo (approved):

tanmayv25 (approved):

eval-dev:
Hi @Tabrizian , One question on thread safety of `at::cuda::setCurrentCUDAStream`. If using one backend thread per model instance mode, different threads can call `at::cuda::setCurrentCUDAStream`, right? Is it a thread safety concern?

eval-dev:
> 

hi @Tabrizian , in your experiences with # of instances from 1 to 4, which execution policy is used? `TRITONBACKEND_EXECUTION_BLOCKING` or `TRITONBACKEND_EXECUTION_DEVICE_BLOCKING`? I wonder which is the best practice when having multiple instances per device. Thanks.

Tabrizian:
You don't have a choice on execution policy when using the PyTorch backend. PyTorch backend always uses `TRITONBACKEND_EXECUTION_BLOCKING`. Also, it would be best to create a new issue since we don't always monitor the comments on pull requests.

PR #77, opened 2022-08-11
Fix stream synchronization

Tabrizian:


GuanLuo (approved):
How much does this explicit synchronization impact performance? Would be less than what we have seen when we introduced the stream execution, right?

Tabrizian:
Yes, it doesn't seem to affect the performance very much:

Before:
```
Concurrency: 16, throughput: 17543.5 infer/sec, latency 908 usec
```

After:
```
Concurrency: 16, throughput: 17093.5 infer/sec, latency 932 usec
```

This if for the `libtorch_device_cpu` model which is an identity model returning outputs on CPU. If the model is performing processing on GPU, the performance is identical to before.

PR #79, opened 2022-08-17
Fix possible "double send" of responses.

GuanLuo:
By this point, the responses are sent but we keep the pointer reference to identify if the request are processed successfully for `TRITONBACKEND_ModelInstanceReportStatistics`. As a result, `RESPOND_ALL_AND_SET_TRUE_IF_ERROR` may send a response twice and cause memory corruption.

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -1223,21 +1223,21 @@ ModelInstanceState::ProcessRequests(
 #ifdef TRITON_ENABLE_GPU
     float compute_input_duration = 0;
     float compute_infer_duration = 0;
-    RESPOND_ALL_AND_SET_TRUE_IF_ERROR(
-        responses, request_count, all_response_failed,
+    LOG_IF_ERROR(
         ConvertCUDAStatusToTritonError(
             cudaEventElapsedTime(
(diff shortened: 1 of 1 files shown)


Tabrizian (approved):

src/libtorch.cc
```diff
@@ -1221,23 +1221,25 @@ ModelInstanceState::ProcessRequests(
   // synchronized the stream in the ReadOutputTensors function.
   if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
 #ifdef TRITON_ENABLE_GPU
+    // [FIXME] in the case of cudaEventElapsedTime failure, should handle
```
Tabrizian, line 1224: I was wondering was this the reason for intermittent segfault after the synchronization has been fixed? `cudaEventElapsedTime` should not fail since we synchronize the stream before reaching this part of the code.

GuanLuo: It's hard to tell, I have not yet been able to catch anything after the synchronization fix (segfault less often?). But note that in general CUDA API may return error from unrelated operation ("Note that this function may also return error codes from previous, asynchronous launches.")

PR #87, opened 2022-12-06
re-enable nvfuser

rmccorm4:
The only thing I notice is this deprecation warning, which I figure we can fix in the future if there's a problem with the old API we're using:
> [W cuda_graph_fuser.h:17] Warning: RegisterCudaFuseGraph::registerPass() is deprecated. Please use torch::jit::fuser::cuda::setEnabled(). (function registerPass)

Fixes https://github.com/triton-inference-server/server/issues/5128

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -432,7 +432,6 @@ ModelState::ParseParameters()
               .c_str());
     }
 
-    // TODO Re-enable NvFuser once fixed
     // If 'ENABLE_NVFUSER' is not present in 'parameters' then no
     // update is made to 'enable_nvfuser'.
     bool enable_nvfuser = false;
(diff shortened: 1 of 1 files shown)


src/libtorch.cc
```diff
@@ -448,8 +447,6 @@ ModelState::ParseParameters()
         TRITONSERVER_ErrorDelete(err);
       }
     } else {
-      // Override, disable NvFuser till fixed
-      enable_nvfuser = false;
       enable_nvfuser_pair_ = {true, enable_nvfuser};
       LOG_MESSAGE(
           TRITONSERVER_LOG_WARN, (std::string("NvFuser is ") +
```
Tabrizian, line 452: Do we need to change this from LOG_WARN to LOG_INFO?

rmccorm4: Sure, changed.

nv-kmcgill53 (approved):

tanmayv25 (approved):

PR #92, opened 2023-02-02
Package linear algebra library

dyastremsky:
Compile linear algebra library into Pytorch backend to be usable with Triton.

Addresses: https://github.com/triton-inference-server/server/issues/4934

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,7 @@ set(PT_LIBS
     "libtorch.so"
     "libtorch_cpu.so"
     "libtorch_cuda.so"
+    "libtorch_cuda_linalg.so"
     "libtorch_global_deps.so"
 )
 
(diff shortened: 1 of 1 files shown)


Tabrizian (commented):
Can you add a test for this too? You can add another model to `L0_libtorch_image_models` that works after adding this library. It would ensure that we have properly linked this library in the backend if anything changes in the future.

dyastremsky:
> Can you add a test for this too? You can add another model to `L0_libtorch_image_models` that works after adding this library. It would ensure that we have properly linked this library in the backend if anything changes in the future.

Makes sense! Do you have a specific reason for suggesting `L0_libtorch_image_models` (e.g. there's an easy way to add an extra image model) or will any test work equally well?

Tabrizian:
I don't have a specific reason. I think that is the place where we have been testing similar functionalities.

dyastremsky:
@Tabrizian Could we submit the testing as a separate PR? I've got it ready, but I need to run it through CI. I would like to get this change merged in time for the release to unblock folks.

Tabrizian (approved):

tanmayv25 (approved):

PR #96, opened 2023-03-29
Fix nvfuser for PyTorch 23.03

mc-nv:


rmccorm4:
@dyastremsky I see it running in latest CI tests, Pytorch tests were moved into their own pipeline now. Thanks for double checking :pray:

dyastremsky:
Ryan shared that this test is in the PyTorch child pipeline, so I'm seeing it now. Thanks, Ryan, this looks to good to merge in!

PR #107, opened 2023-05-17
Add support for instance group of type 'MODEL'

krishung5:
For `MODEL` instance group type,
- Don't specify the device when loading the model so that the backend will not select the device for the model
- For the compute input duration, since only one stream will be used for input collection, we can record the cuda events to calculate the elapsed time. However, for the compute infer duration, since multiple streams will be involved, it is hard to use cuda events as timestamps. Therefore, using `cudaLaunchHostFunc` to launch a callback function to record the timestamp.

Added testing: https://github.com/triton-inference-server/server/pull/5810
Closes: https://github.com/triton-inference-server/server/issues/5694

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -2004,7 +2004,10 @@ ModelInstanceState::SetInputTensors(
 
     // The input must be in contiguous CPU/GPU memory.
     std::vector<std::pair<TRITONSERVER_MemoryType, int64_t>> alloc_perference;
-    if (device_.is_cpu()) {
+    // For 'KIND_MODEL', intput will always be in CPU as we don't have a way to
+    // query the input types.
+    if ((device_.is_cpu()) ||
+        (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL)) {
       alloc_perference = {{TRITONSERVER_MEMORY_CPU_PINNED, 0},
                           {TRITONSERVER_MEMORY_CPU, 0}};
     } else {
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -660,7 +660,14 @@ ModelInstanceState::ModelInstanceState(
 
   if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
 #ifdef TRITON_ENABLE_GPU
-    // Create a CUDA stream for every availble device.
+    // Since we cannot determine the exact devices used by the model, we create
+    // a CUDA stream for every available device to ensure proper synchronization
+    // of CUDA streams. This approach may have implications when a timestamp is
+    // captured on a device that is not used by the model. Currently, this issue
+    // is addressed by synchronizing the CUDA streams before recording
+    // timestamps to prevent timestamp skewing. However, in the future, any
+    // modifications to the CUDA stream synchronization logic should be handled
+    // with caution.
     for (int i = 0; i < torch::cuda::device_count(); i++) {
       cudaStream_t stream;
       THROW_IF_BACKEND_INSTANCE_ERROR(
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -81,6 +81,7 @@ class ModelState : public BackendModel {
   TRITONSERVER_Error* LoadModel(
       const std::string& artifact_name, const torch::Device device,
       std::string* model_path, const TRITONSERVER_InstanceGroupKind& kind,
+      std::unordered_set<int>& device_id_set,
       std::shared_ptr<torch::jit::script::Module>* torch_model);
 
   bool EnabledOptimizedExecution() { return enable_optimized_execution_; }
@@ -206,6 +207,7 @@ TRITONSERVER_Error*
 ModelState::LoadModel(
     const std::string& artifact_name, const torch::Device device,
     std::string* model_path, const TRITONSERVER_InstanceGroupKind& kind,
+    std::unordered_set<int>& device_id_set,
     std::shared_ptr<torch::jit::script::Module>* torch_model)
 {
   // Find the TorchScript file that describes the model. If the model
@@ -256,9 +258,23 @@ ModelState::LoadModel(
   try {
     std::istringstream model_stream(model_data_str);
     if (kind == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
-      // Don't select the device when loading the model.
       torch_model->reset(
           new torch::jit::Module(torch::jit::load(model_stream)));
+
+      // Get the device used in the model
+      auto parameters = (*torch_model)->parameters();
+      auto buffers = (*torch_model)->buffers();
+
+      for (const auto& parameter : parameters) {
+        if (parameter.device().type() != torch::kCPU) {
+          device_id_set.insert(parameter.device().index());
+        }
+      }
+      for (const auto& buffer : buffers) {
+        if (buffer.device().type() != torch::kCPU) {
+          device_id_set.insert(buffer.device().index());
+        }
+      }
     } else {
       torch_model->reset(
           new torch::jit::Module(torch::jit::load(model_stream, device)));
@@ -566,6 +582,13 @@ class ModelInstanceState : public BackendModelInstance {
   cudaEvent_t compute_input_start_event_;
   cudaEvent_t compute_infer_start_event_;
   cudaEvent_t compute_output_start_event_;
+
+  // Store the GPU device ID used in a model for the instance group of type'
+  // MODEL'.
+  std::unordered_set<int> device_id_set_;
+  // Store the extra cuda stream created for the instance group of type' MODEL'
+  // and use device ID as the key.
+  std::unordered_map<int, cudaStream_t> stream_map_;
 };
 
 TRITONSERVER_Error*
@@ -594,10 +617,43 @@ ModelInstanceState::ModelInstanceState(
   if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
 #ifdef TRITON_ENABLE_GPU
     device_ = torch::Device(torch::kCUDA, DeviceId());
+#endif
+  }
+
+  THROW_IF_BACKEND_INSTANCE_ERROR(model_state->LoadModel(
+      ArtifactFilename(), device_, &model_path_, Kind(), device_id_set_,
+      &torch_model_));
+
+#ifdef TRITON_ENABLE_GPU
+  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
+    // Only set the torch device and create a CUDA stream if the model uses GPU.
+    if (!device_id_set_.empty()) {
+      auto it = device_id_set_.begin();
+      // Use the first device to create the default stream.
+      THROW_IF_BACKEND_INSTANCE_ERROR(
+          CreateCudaStream(*it, 0 /* cuda_stream_priority */, &stream_));
+      device_ = torch::Device(torch::kCUDA, *it);
+
+      // Create a CUDA stream for other devices so that they can be synchronized
+      // later. Skip the first device since it is used to create the default
+      // stream.
+      if (it != device_id_set_.end()) {
+        ++it;
+      }
+      for (; it != device_id_set_.end(); ++it) {
+        cudaStream_t stream;
+        THROW_IF_BACKEND_INSTANCE_ERROR(
+            CreateCudaStream(*it, 0 /* cuda_stream_priority */, &stream));
+        stream_map_.insert({*it, stream});
+      }
+    }
+  }
+
+  if (device_.is_cuda()) {
     // Need to set the CUDA context so that the context that events are
     // created on match with contexts that events are recorded with.
     THROW_IF_BACKEND_INSTANCE_ERROR(ConvertCUDAStatusToTritonError(
-        cudaSetDevice(DeviceId()), TRITONSERVER_ERROR_INTERNAL,
+        cudaSetDevice(device_.index()), TRITONSERVER_ERROR_INTERNAL,
         "Failed to set the device"));
     THROW_IF_BACKEND_INSTANCE_ERROR(ConvertCUDAStatusToTritonError(
         cudaEventCreate(&compute_input_start_event_),
@@ -608,11 +664,8 @@ ModelInstanceState::ModelInstanceState(
     THROW_IF_BACKEND_INSTANCE_ERROR(ConvertCUDAStatusToTritonError(
         cudaEventCreate(&compute_output_start_event_),
         TRITONSERVER_ERROR_INTERNAL, "Failed to create cuda event"));
-#endif
   }
-
-  THROW_IF_BACKEND_INSTANCE_ERROR(model_state->LoadModel(
-      ArtifactFilename(), device_, &model_path_, Kind(), &torch_model_));
+#endif
 
   size_t expected_input_cnt = 0;
   {
@@ -680,6 +733,21 @@ ModelInstanceState::~ModelInstanceState()
 {
   torch_model_.reset();
   ClearCache();
+
+  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
+    for (auto& m : stream_map_) {
+      cudaSetDevice(m.first);
+      cudaError_t err = cudaStreamDestroy(m.second);
+      if (err != cudaSuccess) {
+        TRITONSERVER_LogMessage(
+            TRITONSERVER_LOG_ERROR, __FILE__, __LINE__,
+            (std::string("~ModelInstanceState: ") + name_ +
+             " failed to destroy cuda stream: " + cudaGetErrorString(err))
+                .c_str());
+      }
+      m.second = nullptr;
+    }
+  }
 }
 
 TRITONSERVER_Error*
@@ -1039,13 +1107,16 @@ ModelInstanceState::ProcessRequests(
        std::to_string(request_count) + " requests")
           .c_str());
 
-  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
 #ifdef TRITON_ENABLE_GPU
-    at::cuda::CUDAStream torch_stream =
-        at::cuda::getStreamFromExternal(stream_, DeviceId());
+  if ((Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) ||
+      (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL && device_.is_cuda())) {
+    at::cuda::CUDAStream torch_stream = at::cuda::getStreamFromExternal(
+        stream_, (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU)
+                     ? DeviceId()
+                     : device_.index());
     at::cuda::setCurrentCUDAStream(torch_stream);
-#endif
   }
+#endif
 
   NVTX_RANGE(nvtx_, "ProcessRequests " + Name());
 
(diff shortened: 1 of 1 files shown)


src/libtorch.cc
```diff
@@ -581,10 +610,43 @@ ModelInstanceState::ModelInstanceState(
...
+      auto it = device_id_set_.begin();
+      // Use the first device to create the default stream.
+      THROW_IF_BACKEND_INSTANCE_ERROR(
+          CreateCudaStream(*it, 0 /* cuda_stream_priority */, &stream_));
+      device_ = torch::Device(torch::kCUDA, *it);
+
+      // Create a CUDA stream for other devices so that they can be synchronized
+      // later. Skip the first device since it is used to create the default
+      // stream.
+      if (it != device_id_set_.end()) {
```
nnshah1, line 633: Small nit: this can't be false - as we dereference it above. If it == device_id.set_.end() then *it would cause an exception.
We can put the ++it in the for loop below:
```
for (++it; it!=device_id_set.end();++it) {
```

src/libtorch.cc
```diff
@@ -253,8 +255,28 @@ ModelState::LoadModel(
...
-        new torch::jit::Module(torch::jit::load(model_stream, device)));
+    if (kind == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
+      torch_model->reset(
+          new torch::jit::Module(torch::jit::load(model_stream)));
+
+      // Get the device used in the model
+      auto parameters = (*torch_model)->parameters();
+      auto buffers = (*torch_model)->buffers();
+
+      for (const auto& parameter : parameters) {
```
tanmayv25, line 266: Obtaining device_id sets can be simplified to a templated utility function that can consume both parameters and buffers data type.
Should be able reduce the code duplicity.

src/libtorch.cc
```diff
@@ -1000,13 +1074,16 @@ ModelInstanceState::ProcessRequests(
        std::to_string(request_count) + " requests")
           .c_str());
 
-  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
 #ifdef TRITON_ENABLE_GPU
-    at::cuda::CUDAStream torch_stream =
-        at::cuda::getStreamFromExternal(stream_, DeviceId());
+  if ((Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) ||
+      (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL && device_.is_cuda())) {
+    at::cuda::CUDAStream torch_stream = at::cuda::getStreamFromExternal(
```
tanmayv25, line 1080: Is calling `setCurrentCUDAStream` to the cuda stream corresponding to the first device, really the right thing to do for the KIND_MODEL. Is setting the current cuda stream even required for kind_model?

src/libtorch.cc
```diff
@@ -253,8 +255,28 @@ ModelState::LoadModel(
...
   try {
     std::istringstream model_stream(model_data_str);
-    torch_model->reset(
-        new torch::jit::Module(torch::jit::load(model_stream, device)));
+    if (kind == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
+      torch_model->reset(
+          new torch::jit::Module(torch::jit::load(model_stream)));
+
+      // Get the device used in the model
+      auto parameters = (*torch_model)->parameters();
```
Tabrizian, line 263: There is also the `attributes` field: https://pytorch.org/cppdocs/api/structtorch_1_1jit_1_1_module.html#_CPPv4NK5torch3jit6Module10attributesEb

Do we need to include that in the list hear too?

Tabrizian (changes requested):
Could you please explain how this would solve the cuda synchronization issue? 

Just creating the streams doesn't mean that PyTorch would use them. Looks like you are only creating streams for other devices but they are not actually being used by PyTorch for the execution of the model.

src/libtorch.cc
krishung5: Discussed offline that we should call `setCurrentCUDAStream` for the cuda stream of every device to replace the default stream.

src/libtorch.cc
krishung5: Nice catch! Missed the `attributes` field.
However, there seems to be no proper way to retrieve all the device used in a model based on the [comment](https://github.com/pytorch/pytorch/issues/7460#issuecomment-388013922). Although we can get the attributes, parameters and the buffers from a model, it is possible that the model contains complex structures or custom-defined objects where the device information may not be directly accessible. Additionally, it's challenging to test every corner case as we may not have access to all possible models. If a device id is missed, it can lead to data corruption because the corresponding CUDA stream is not synchronized. To address this, I was thinking we could create a new CUDA stream for every available device. This ensures that each stream is properly synchronized. Let me know your thoughts on this.

src/libtorch.cc
```diff
@@ -581,10 +610,43 @@ ModelInstanceState::ModelInstanceState(
...
+  }
+
+  THROW_IF_BACKEND_INSTANCE_ERROR(model_state->LoadModel(
+      ArtifactFilename(), device_, &model_path_, Kind(), device_id_set_,
+      &torch_model_));
+
+#ifdef TRITON_ENABLE_GPU
+  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
+    // Only set the torch device and create a CUDA stream if the model uses GPU.
+    if (!device_id_set_.empty()) {
```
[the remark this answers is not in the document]
krishung5: Makes sense! However, I don' think we need set the `device_` anymore so this part has been removed.

krishung5:
Documentation added in a separate PR: https://github.com/triton-inference-server/pytorch_backend/pull/110

src/libtorch.cc
```diff
@@ -59,6 +59,107 @@
...
+  TimestampData()
+      : compute_input_start_(0), earliest_compute_infer_start_(0),
+        last_compute_infer_start_(0), compute_output_start_(0)
+  {
+  }
+
+  enum struct TimestampHint { COMPUTE_INPUT, COMPUTE_INFER, COMPUTE_OUTPUT };
+
+  // Set the timestamp hint
+  void SetTimestampHint(const TimestampHint& hint) { hint_ = hint; }
```
Tabrizian, line 77: Instead of setting the hint separately how about making it an argument to `CaptureTimestamp`?

Tabrizian (commented):
Do we need to update this part too? https://github.com/triton-inference-server/pytorch_backend/blob/ead0e23b432799ca6d0c13d6dfbe2fab904f96e0/src/libtorch.cc#L1774-L1779

I guess for KIND_MODEL the inputs will be always in CPU since we don't have a way to query the input types required by the model beforehand?

src/libtorch.cc
krishung5: The `CaptureTimestamp` function is called inside the callback function where only one argument can be passed: https://github.com/triton-inference-server/pytorch_backend/pull/107/files#diff-859dcc1d94f2d13b0d19e7a14af430cd18eb66395a6f69f2d4886854319c4467R154. We could pass both `timestamp_data_` and `hint` as a tuple to the callback but I thought using a separate function makes the code more straightforward.

krishung5:
> Do we need to update this part too?
> 
> https://github.com/triton-inference-server/pytorch_backend/blob/ead0e23b432799ca6d0c13d6dfbe2fab904f96e0/src/libtorch.cc#L1774-L1779
> 
> I guess for KIND_MODEL the inputs will be always in CPU since we don't have a way to query the input types required by the model beforehand?

That is correct. Here the `device_.is_cpu()` will be true for `KIND_MODEL` because it is initialized with [torch::kCPU](https://github.com/triton-inference-server/pytorch_backend/pull/107/files#diff-859dcc1d94f2d13b0d19e7a14af430cd18eb66395a6f69f2d4886854319c4467R705) and is not updated to any GPU device for `KIND_MODEL`. However, this condition can be indirect, so I have updated the if condition and the comment to make it clearer.

src/libtorch.cc
```diff
@@ -1341,8 +1523,9 @@ ModelInstanceState::Execute(
 
     // NV-Fuser. No change is made unless parameter is explicitly set.
     if (std::get<0>(model_state_->EnabledNvfuserPair())) {
-      if (std::get<1>(model_state_->EnabledNvfuserPair()) &&
-          (device_.type() != torch::kCPU)) {
+      if ((std::get<1>(model_state_->EnabledNvfuserPair()) &&
+           (device_.type() != torch::kCPU)) ||
+          (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL)) {
```
Tabrizian, line 1528: Would this work if the model is loaded with KIND_MODEL but runs on CPU and the container is loaded without --gpus flag?

krishung5: Good catch. I think we need to add the check for available devices here.

src/libtorch.cc
```diff
@@ -665,7 +796,7 @@ void
 ModelInstanceState::ClearCache()
 {
 #ifdef TRITON_ENABLE_GPU
-  if (device_.is_cuda()) {
+  if ((device_.is_cuda()) || (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL)) {
```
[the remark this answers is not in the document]
krishung5: I think we need to add a check here and only call `emptyCache` below if there are available devices.

src/libtorch.cc
```diff
@@ -62,6 +62,107 @@
...
+  // Get the earliest timestamp.
+  void GetEarliestTimestamp(uint64_t& timestamp);
+
+  // Get the latest timestamp.
+  void GetLastTimestamp(uint64_t& timestamp);
+
+  // Get the duration of comput input.
+  uint64_t GetComputeInputDuration()
+  {
+    return last_compute_infer_start_ - compute_input_start_;
```
tanmayv25, line 94: For multi-device inference runs, wouldn't it be better to focus on capturing the complete compute_infer times? 
We should aim for `compute_input + compute_infer + compute_output` ==  `total_compute`?

I believe it would be better to capture: compute_input_start_, first_compute_infer_start_, last_compute_infer_end_.
The duration can be captured as:
compute_input = first_compute_infer_start_ - compute_input_start_
compute_infer = last_compute_infer_end_ - first_compute_infer_start_
compute_output = compute_output_end(last) - last_compute_infer_end_

I know this is not 100% accurate as a part of compute_input and compute_output might seep into compute_infer, but it is more comprehensible. 
@Tabrizian @krishung5 @nnshah1  what do you think?

src/libtorch.cc
tanmayv25: How about having two variants of `TimestampCaptureCallback` function in this class? One with timestamp override to capture the last timestamp and the other without timestamp override to capture the first timestamp?

src/libtorch.cc
tanmayv25: `first_compute_infer_start_` should not be that tricky as we synchronize all the streams before capturing this timestamp. Hence, most of these streams should capture identical timestamp. The `first_compute_infer_start_` will be the tricky one.

src/libtorch.cc
```diff
@@ -607,7 +726,19 @@ ModelInstanceState::ModelInstanceState(
   }
 
   THROW_IF_BACKEND_INSTANCE_ERROR(model_state->LoadModel(
-      ArtifactFilename(), device_, &model_path_, &torch_model_));
+      ArtifactFilename(), device_, &model_path_, Kind(), &torch_model_));
+
+  if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
+#ifdef TRITON_ENABLE_GPU
+    // Create a CUDA stream for every availble device.
```
tanmayv25, line 733: Is there no way to capture what cuda devices are used by this model instance?
It's good that we are synchronizing against all these streams before capturing first_compute_infer_start_ event, otherwise the timestamps would have been skewed. You should document this implication of creating streams and events on all available devices. So, that in future we are careful when working with this code. 
For example, imagine if someone is capturing first_compute_output_start_ event on a device that isn't running the model.

src/libtorch.cc
```diff
@@ -1153,6 +1306,16 @@ ModelInstanceState::ProcessRequests(
         ConvertCUDAStatusToTritonError(
             cudaEventRecord(compute_input_start_event_, stream_),
             TRITONSERVER_ERROR_INTERNAL, "Failed to record the event."));
+#endif
+  } else if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
+#ifdef TRITON_ENABLE_GPU
+    timestamp_data_.SetTimestampHint(
+        TimestampData::TimestampHint::COMPUTE_INPUT);
+    for (const auto& stream : stream_vec_) {
+      cudaLaunchHostFunc(
```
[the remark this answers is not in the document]
Tabrizian: I see. I guess there could be a way by using cudaStreamWaitEvent but I think what you have is easier to implement/understand. Resolving.

src/libtorch.cc
krishung5: Fixed the condition.

src/libtorch.cc
krishung5: Used two callback function to capture the first the the last timestamp.

src/libtorch.cc
krishung5: Added comment in this [commit](https://github.com/triton-inference-server/pytorch_backend/commit/65e8cb49481903f0ef33211fbbcc4576459e27d1). Not sure why it's not showing on the PR..

src/libtorch.cc
krishung5: With adding cuda stream synchronization before reading the outputs, we wouldn't need the `first_compute_infer_start_` and `last_compute_infer_end_`.

src/libtorch.cc
```diff
@@ -1146,21 +1238,36 @@ ModelInstanceState::ProcessRequests(
   std::vector<torch::jit::IValue> input_tensors;
   bool cuda_copy = false;
   std::unique_ptr<BackendInputCollector> collector;
+  std::mutex timestamp_mu;
+
+  uint64_t compute_input_start = 0;
```
Tabrizian, line 1243: Instead of mutex / variable combination you can use an atomic variable: std::atomic<uint64_t>

src/libtorch.cc
```diff
@@ -1146,21 +1238,36 @@ ModelInstanceState::ProcessRequests(
...
             cudaEventRecord(compute_input_start_event_, stream_),
             TRITONSERVER_ERROR_INTERNAL, "Failed to record the event."));
+#endif
+  } else if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) {
+#ifdef TRITON_ENABLE_GPU
+    for (const auto& stream : stream_vec_) {
+      cudaLaunchHostFunc(
+          stream, CaptureFirstTimestampCallback,
+          reinterpret_cast<void*>(&compute_input_cb_data));
+    }
```
Tabrizian, line 1261: As discussed, I don't think for inputs we need to make any changes. We only need the multi-stream logic for calculating the `compute infer` time.

src/libtorch.cc
```diff
@@ -1170,19 +1277,23 @@ ModelInstanceState::ProcessRequests(
...
     cuda_copy = false;
   }
 #endif
 
   std::vector<torch::jit::IValue> output_tensors;
   uint64_t compute_start_ns = 0;
+  uint64_t compute_infer_start = 0;
 
+  std::tuple<uint64_t*, std::mutex*> compute_infer_cb_data(
+      &compute_infer_start, &timestamp_mu);
```
Tabrizian, line 1290: I don't think these changes are required.

src/libtorch.cc
krishung5: Resolving as it is no longer relevant in the new changes.

src/libtorch.cc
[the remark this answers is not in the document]
krishung5: Resolving as it's no longer relevant in the new changes.

src/libtorch.cc
krishung5: Updated to use std::atomic<uint64_t> and removed the mutex.

src/libtorch.cc
krishung5: Removed this part and use cuda events for the comput input duration.

src/libtorch.cc
krishung5: We are using cuda events for the comput input duration, but we would still need the `compute_infer_start` timestamp for the compute infer duration.

krishung5:
@Tabrizian Addressed all the comments from the offline discussion. Please review, thank you!

src/libtorch.cc
```diff
@@ -1213,11 +1305,31 @@ ModelInstanceState::ProcessRequests(
...
 
   uint64_t compute_end_ns = 0;
-  RESPOND_ALL_AND_SET_TRUE_IF_ERROR(
-      responses, request_count, all_response_failed,
-      RecordBackendTimestamp(
-          &compute_end_ns,
-          reinterpret_cast<void*>(&compute_output_start_event_)));
+  std::atomic<uint64_t> compute_output_start = 0;
+
+  if ((Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) && (device_cnt_ > 0)) {
```
Tabrizian, line 1310: I think for KIND_MODEL and device_cnt == 0 the stats are not handled properly.

src/libtorch.cc
```diff
@@ -1249,28 +1361,24 @@ ModelInstanceState::ProcessRequests(
...
+    float compute_input_duration = GetCudaEventElapsedTime(
+        compute_input_start_event_, compute_infer_start_event_);
+    float compute_infer_duration = GetCudaEventElapsedTime(
+        compute_infer_start_event_, compute_output_start_event_);
 
     compute_start_ns = exec_start_ns + (compute_input_duration * 1e6);
     compute_end_ns = compute_start_ns + (compute_infer_duration * 1e6);
+#endif
+  } else if (
+      (Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) && (device_cnt_ > 0)) {
```
Tabrizian, line 1373: what about KIND_MODEL and device_cnt_ ==0 ? It looks like it is not handled.

src/libtorch.cc
```diff
@@ -2049,19 +2172,88 @@ TRITONSERVER_Error*
...
+  // [FIXME] in the case of cudaEventElapsedTime failure, should handle
+  // stats reporting more gracefully as the durations are inaccurate
+  float duration = 0;
+  LOG_IF_ERROR(
+      ConvertCUDAStatusToTritonError(
+          cudaEventElapsedTime(&duration, start_event, end_event),
+          TRITONSERVER_ERROR_INTERNAL, "Failed to capture elapsed time"),
+      "Failed to capture elapsed time");
+  return duration;
+#endif
```
Tabrizian, line 2254: I think you need to return 0 for if the GPU is not enabled so that it compiles successfully.

src/libtorch.cc
krishung5: I think for KIND_MODEL and device_cnt_ ==0, we won't need to calculate the `compute_start_ns` and `compute_end_ns` here. The `compute_start_ns` and `compute_end_ns` is already captured in `RecordBackendTimestamp` and will be used as is, just like the KIND_CPU case.

src/libtorch.cc
krishung5: I think for KIND_MODEL and device_cnt == 0, it will just go to the else case and capture the `compute_end_ns` in `RecordBackendTimestamp`.

src/libtorch.cc
krishung5: Good catch, fixed.

src/libtorch.cc
Tabrizian: sorry, I think I now understand how it is handled. Resolving.

Tabrizian (dismissed):

krishung5:
Fix up the initialization of `std::atomic` which was failing on jetson build.

Tabrizian (dismissed):

tanmayv25 (changes requested):

src/libtorch.cc
```diff
@@ -1213,11 +1305,31 @@ ModelInstanceState::ProcessRequests(
...
-  RESPOND_ALL_AND_SET_TRUE_IF_ERROR(
-      responses, request_count, all_response_failed,
-      RecordBackendTimestamp(
-          &compute_end_ns,
-          reinterpret_cast<void*>(&compute_output_start_event_)));
+  std::atomic<uint64_t> compute_output_start{0};
+
+  if ((Kind() == TRITONSERVER_INSTANCEGROUPKIND_MODEL) && (device_cnt_ > 0)) {
+#ifdef TRITON_ENABLE_GPU
+    // For the compute infer duration, multiple streams will be involved, so we
```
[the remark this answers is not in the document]
krishung5: Updated the code, please review.

tanmayv25 (approved):

PR #108, opened 2023-05-22
Setting max_gpu_fraction as in Tensorflow backend

FabianSchuetze:
This PR permits setting the a max_gpu_fraction for the pytorch backend. 

Pytorch allows setting the max gpu fraction through the [`CUDACachingAllocator`](https://github.com/pytorch/pytorch/blob/main/c10/cuda/CUDACachingAllocator.h) . The user of the pytorch_backend can set the memory fraction in the same fashion as in the [tensorflow backend](https://github.com/triton-inference-server/tensorflow_backend/blob/main/src/tensorflow.cc#L2636). The memory fraction applies to all models. 

I am a bit uncertain about how to handle the case with multiple GPUs and would appreciate feedback about desired behavior in this case.

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -25,7 +25,9 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <stdint.h>
+
 #include <exception>
+
 #include "libtorch_utils.h"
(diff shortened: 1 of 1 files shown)


whateverforever:
any update on this?

FabianSchuetze:
I haven't heard from the owners of the repo yet. Is someone available for a code review?

PR #110, opened 2023-06-05
Add documentation for instance group kind of type 'KIND_MODEL'

krishung5:
The documentation for https://github.com/triton-inference-server/pytorch_backend/pull/107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,13 +208,25 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
 
 ### Support 
 
-Starting from the 23.06 release, the PyTorch backend supports an instance group
-kind of type
-[`KIND_MODEL`](https://github.com/triton-inference-server/common/blob/r23.05/protobuf/model_config.proto#L174-L181)
-where the backend will not choose the GPU device for the model. Instead, it
-will respect the device(s) used in the model and use it as is when the type of
-the instance group is set to `KIND_MODEL` in the model config file. This is
-useful when the model is using multiple GPUs internally.
+#### Model Instance Group Kind
+
+The PyTorch backend supports the following kinds of
+[Model Instance Groups](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_configuration.md#instance-groups)
+where the input tensors are placed as follows:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -206,6 +206,16 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
 
     `ENABLE_TENSOR_FUSER`
 
+### Support
```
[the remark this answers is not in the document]
krishung5: Great suggestion. Added the description for inputs to the doc.

README.md
```diff
@@ -206,6 +206,28 @@ complex execution modes and dynamic shapes. If not specified, all are enabled by
...
+instance.
+
+* `KIND_CPU`: Inputs are on the CPU.
+
+* `KIND_MODEL`:  Starting from the 23.06 release, the PyTorch backend supports
+instance group kind of type
+[`KIND_MODEL`](https://github.com/triton-inference-server/common/blob/r23.05/protobuf/model_config.proto#L174-L181).
+In this case, the inputs reside on the CPU. The backend does not choose the GPU
+device for the model; instead, it respects the device(s) specified in the model
+and uses them as they are when the instance group kind is set to `KIND_MODEL`
```
dyastremsky, line 227: Nit: I think this section could be shortened, which could add clarity. It feels like a bit of a run on. I think something like the previous sections could be good, like "Inputs are located on the devices specified by the model. This feature is available starting in the 23.06 release."

Also, a couple of questions:
- Is there a default case when the model does not specify a device? Does this fail? If not, wondering if the default value is worth adding here.
- Is there a test model that we could link to? It's okay if not. If we do already have it, linking to that as an example could be useful. It would show users how to specify devices in a PyTorch model.

krishung5: Modified the section, let me know if there is anything unclear!
> Is there a default case when the model does not specify a device? Does this fail? If not, wondering if the default value is worth adding here.

By default the first available GPU device will be used. Added this to the documentation.

>Is there a test model that we could link to? It's okay if not. If we do already have it, linking to that as an example could be useful. It would show users how to specify devices in a PyTorch model.

Added the link to the testing model.

dyastremsky (approved):

Tabrizian (approved):

PR #113, opened 2023-06-08
Copy HPCX paths for CPU images

dyastremsky:
Copy HPCX paths from the PyTorch container for CPU-only builds, so that it has the expected version of HPCX dependencies for libtorch.so.

Server: https://github.com/triton-inference-server/server/pull/5922

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,14 +202,14 @@ if(NOT TRITON_ENABLE_GPU)
     OUTPUT
       ${HPCX_LIBS}
     COMMAND docker pull ${TRITON_PYTORCH_DOCKER_IMAGE}
-    COMMAND docker rm pytorch_backend_ptlib || echo "error ignored..." || true
-    COMMAND docker create --name pytorch_backend_ptlib ${TRITON_PYTORCH_DOCKER_IMAGE}
-    COMMAND docker cp -L pytorch_backend_ptlib:/opt/hpcx/ucc/lib/libucc.so.1 libucc.so.1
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -189,6 +189,31 @@ set(OPENCV_LIBS
...
+    COMMAND docker pull ${TRITON_PYTORCH_DOCKER_IMAGE}
+    COMMAND docker rm pytorch_backend_hpcxlib || echo "error ignored..." || true
+    COMMAND docker create --name pytorch_backend_hpcxlib ${TRITON_PYTORCH_DOCKER_IMAGE}
+    COMMAND docker cp -L pytorch_backend_hpcxlib:/opt/hpcx/ucc/lib/libucc.so.1 libucc.so.1
+    COMMAND docker cp -L pytorch_backend_hpcxlib:/opt/hpcx/ucx/lib/libucm.so.0 libucm.so.0
+    COMMAND docker cp -L pytorch_backend_hpcxlib:/opt/hpcx/ucx/lib/libucp.so.0 libucp.so.0
+    COMMAND docker cp -L pytorch_backend_hpcxlib:/opt/hpcx/ucx/lib/libucs.so.0 libucs.so.0
+    COMMAND docker cp -L pytorch_backend_hpcxlib:/opt/hpcx/ucx/lib/libuct.so.0 libuct.so.0
+    COMMAND docker rm pytorch_backend_hpcxlib
+    VERBATIM
```
Tabrizian, line 213: Instead of modifying cmake can we copy over the whole `/opt/hpcx/ucc/` and `/opt/hpcx/ucx` in the build.py?

dyastremsky: Sure, I will try this approach.

PR #117, opened 2023-11-14
Bring back Python backend based PyTorch backend

kthui:
Related PRs:
* https://github.com/triton-inference-server/server/pull/6518
* https://github.com/triton-inference-server/core/pull/285
* https://github.com/triton-inference-server/common/pull/103
* https://github.com/triton-inference-server/backend/pull/94

Add Python backend based PyTorch runtime. Add script to build Conda-Pack environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ model_repository/
 
 The `model.pt` is the TorchScript model file.
 
-### Customization
+## Customization
 
 The following PyTorch settings may be customized by setting parameters on the
 `config.pbtxt`.


src/model.py
```diff
@@ -0,0 +1,318 @@
...
+
+# triton_python_backend_utils is available in every Triton Python model. You
+# need to use this module to create inference requests and responses. It also
+# contains some utility functions for extracting information from model_config
+# and converting Triton input/output types to numpy types.
+import triton_python_backend_utils as pb_utils
+
+
+def _get_model_path(config):
+    filenames = ["model.py", "model.pt"]
```
rmccorm4, line 46: If I am understanding the whole flow correctly:

1. The `runtime` flag for python based implementation will look for `model.py` in model/version repo, and then backend repo (in that order)
2. The python runtime for pytorch backend supports loading pytorch NN module style models as a `model.py` file

Since the core looks in model repo first, if you tried to load a pytorch `models/resnet50/1/model.py` (nn module file) that uses the Triton python runtime pytorch implementation (`/opt/tritonserver/backends/pytorch/model.py`) -- Triton core would probably try to load this `models/resnet50/1/model.py` as the backend implementation and fail, right?

kthui: I guess this might be the most confusing part. If a model wants to load on top of Python backend, the `backend_libdir` and `backend_libpath` will initially point to the C++ runtime of the Python backend. After that, the `backend_libdir` will be "updated" to point to the Python runtime (i.e. `/opt/tritonserver/backends/pytorch/model.py`), without modifying `backend_libpath`.

The reason why the "update" logic is always correct is the Python runtime path is always assembled from `backend_dir` and `model.py`, where `backend_dir` always points to `/opt/tritonserver/backends/pytorch`, so the eventual `backend_libdir` can only be `/opt/tritonserver/backends/pytorch/model.py` for example.

kthui: I think we should rename the runtime from `model.py` to `backend.py` to avoid any ambiguity and allow for better search logic, but `model.py` is used since TRT-LLM, so it will be a behavioral change if we do so.

kthui: [Limit Python based backend search to backend directory](https://github.com/triton-inference-server/core/pull/285/commits/d2c456f7ba312f87da4b3dd0740643dd2a23042c)

tools/gen_pb_exec_env.sh
```diff
@@ -0,0 +1,52 @@
...
+conda create -n pt python=3.10 -y
+conda activate pt
+conda install -c conda-forge conda-pack -y
+
+# pre install step
+export PYTHONNOUSERSITE=True
+conda install -c conda-forge libstdcxx-ng=12 -y
+
+# install PyTorch
+conda install pytorch torchvision torchaudio pytorch-cuda=12.1 -c pytorch -c nvidia -y
```
oandreeva-nv, line 44: if we don't specify version `12.1` for pytorch-cuda, will it still install the latest stable version?

src/model.py
oandreeva-nv, line 46: do we want to support `<XXX>.pt2` ? i.e result of `torch.export` : https://pytorch.org/docs/stable/export.html#serialization

kthui: How about we add `.pt2` support as a follow-up ticket? It is because I think it depends on `torch.export` which is not part of the previous "platform handler", so we do not introduce additional risks for this "bring back platform handler" ticket and get some functionality in sooner than later. (unless `.pt2` requires behavioral changes if introduced separately?)

oandreeva-nv: Works for me. Can we add this note as a FIXME? relevant ticket is: https://jirasw.nvidia.com/browse/DLIS-5694

kthui: Issue is addressed, with commit up to https://github.com/triton-inference-server/core/pull/285/commits/f502bfc5f8c2c50d8e170a09b776ffd2f6d0d1f6 point

kthui: [Add note for adding .pt2 model support](https://github.com/triton-inference-server/pytorch_backend/pull/117/commits/4381340963e20f1da8da395ff0df77e1ca292602)

tools/gen_pb_exec_env.sh
kthui: Without specifying version `12.1` for pytorch-cuda worked locally. I am ok with removing the specified version. My only concern is why PyTorch suggests setting `pytorch-cuda=12.1` at the first place:
```
conda install pytorch torchvision torchaudio pytorch-cuda=12.1 -c pytorch -c nvidia
```
Would it break in the future?
https://pytorch.org/get-started/locally/

kthui: [Do not specify pytorch cuda version](https://github.com/triton-inference-server/pytorch_backend/pull/117/commits/f71dd1775731a8049060225f183027ab02a8d0b9)

CMakeLists.txt
```diff
@@ -504,6 +505,21 @@ install(
     ${INSTALL_CONFIGDIR}
 )
 
+if (${TRITON_PYTORCH_ENABLE_PYTHON_RUNTIME})
```
Tabrizian, line 508: Does this mean that we'll include the runtime by default? I believe the environment size could be significant since it contains CUDA, PyTorch, etc.

src/model.py
```diff
@@ -0,0 +1,319 @@
...
+
+try:
+    import torch
+except ModuleNotFoundError as error:
+    raise RuntimeError("Missing/Incomplete PyTorch package installation") from error
+
+# triton_python_backend_utils is available in every Triton Python model. You
+# need to use this module to create inference requests and responses. It also
+# contains some utility functions for extracting information from model_config
+# and converting Triton input/output types to numpy types.
```
Tabrizian, line 41: Let's remove the comment.

CMakeLists.txt
kthui: Yes, but the behavior can be easily changed. The current size of the conda pack is 3.00 GB.

src/model.py
kthui: Removed: [Remove legacy comment](https://github.com/triton-inference-server/pytorch_backend/pull/117/commits/78c47fe8b5901ba42283cba41702bc19bae18500)

README.md
```diff
@@ -243,3 +243,112 @@ instance in the
...
+
+## Model Layout
+
+The model repository should look like:
+
+```
+model_repository/
+`-- model_directory
+    |-- 1
+    |   |-- model.py
```
Tabrizian, line 296: I think this is probably a little bit misleading since we don't always require both `model.py` and `model.pt` (e.g., if torchscript is provided only `model.pt` is required as you mentioned).

kthui: Good spotting! I created different sections for PyTorch 2.0 and TorchScript model layout. [Clarify model layout between PyTorch and TorchScript](https://github.com/triton-inference-server/pytorch_backend/pull/117/commits/b8abcaa12df3d603e99e339c5b59f1c5ffc476b6)

src/model.py
rmccorm4: Try to avoid force pushing if possible, as the [commit referenced previously](https://github.com/triton-inference-server/pytorch_backend/pull/117/files#r1407007124) is no longer valid.

rmccorm4: Can you clarify the current behavior now? 

For python-based backends, it will only look for `model.py` in the backend directory.

For C++ backends, will it still look in (version_dir, model_dir, backend_dir)?

rmccorm4 (commented):
This PR generally looks good to me - just looking for some clarifications to close this thread: https://github.com/triton-inference-server/pytorch_backend/pull/117/files#r1447954567

src/model.py
kthui: > For python-based backends, it will only look for model.py in the backend directory.
For C++ backends, will it still look in (version_dir, model_dir, backend_dir)?

Yes.

> how does this interact with escape logic?

The `backend_libdir` will be one of the (version_dir, model_dir, backend_dir) path. The `runtime` is essentially the `backend_libname`, which `<backend_libdir>/<runtime>` is the `backend_libpath`. At the end after everything are determined, there is a one time check to ensure the `backend_libpath` is within the `backend_libdir`. If the `runtime` tries to escape from the `backend_libdir` (i.e. `runtime = "../my_backend_lib.so" -> backend_libpath = "backend_dir/../my_backend_lib.so"`), this will be caught.

> can you verify running a custom backend with shared library located in the model folder still works?

Yes. I think the [L0_lifecycle](https://github.com/triton-inference-server/server/blob/main/qa/L0_lifecycle/test.sh#L1258) covers such case, and it is part of the CI run. It copies the `libtriton_identity.so` into the model folder, see `cp libtriton_identity.so models/identity_zero_1_int32/1/.` line.

kthui: I also ran L0_passive_instance on the test container from CI and it passed, which uses a custom C++ backend: https://github.com/triton-inference-server/server/blob/main/qa/L0_passive_instance/models/distributed_int32_int32_int32/config.pbtxt#L28

rmccorm4 (approved):

PR #120, opened 2023-12-12
fix the crashing when there are zero-size inputs

yongbinfeng:
PR to fix the crashing when some of the PyTorch inputs have zero size. One example here:

https://github.com/triton-inference-server/server/issues/5972

The error seems tbe be because of `Torch::from_blob` function for copying inputs not working when the tensor is zero-size. The fix is to add the special handle in this case, directly creating a zero-size tensor, instead of calling the PyTorch function.

Tested and it seems to have fixed the crashing.


Tabrizian (approved):
I think I just need to add some testing and run it through the CI otherwise it is looking good.

Tabrizian (approved):

PR #125, opened 2024-03-06
add thread control for pytorch backend

yongbinfeng:
As noted in the issue here: https://github.com/triton-inference-server/server/issues/6896 we have found out the number of threads can affect the pytorch inference performance a lot. In some cases we have seen PyTorch inference runs (super) slow on multi-core CPU machines, and just setting number of instance is not enough to handle the problem. We have tested using `at::set_num_threads(1)` and confirmed this fixes the slow inference issue.

This PR allows to configure `intra_op_thread_count` and `inter_op_thread_count` for pytorch models, similar to other backends such as TF and ONNX, with syntax such as

```
parameters { key: "INTRA_OP_THREAD_COUNT" value: { string_value: "1" } }
parameters { key: "INTER_OP_THREAD_COUNT" value: { string_value: "1" } }
```

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -56,6 +56,12 @@
 #include <cuda_runtime_api.h>
 #endif  // TRITON_ENABLE_GPU
 
+// for thread control
+// https://pytorch.org/docs/stable/notes/cpu_threading_torchscript_inference.html#runtime-api
+// https://github.com/pytorch/pytorch/blob/v2.2.1-rc3/aten/src/ATen/Parallel.h#L133
+#include <ATen/Parallel.h>
+
+
 //
 // PyTorch C++ (LibTorch) Backend that implements the TRITONBACKEND API.
 //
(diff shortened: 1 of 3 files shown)


pascualex:
Thank you for this, we are experiencing the same problem and are being forced to convert our models to ONNX, which in turn is causing other issues.

I'm not a maintainer so I can only validate that this is a real issue for us too.

src/libtorch.cc
```diff
@@ -465,6 +471,54 @@ ModelState::ParseParameters()
            " for model instance '" + Name() + "'")
               .c_str());
     }
+
+    // If "INTRA_OP_THREAD_COUNT" is not present in 'parameters' then no update
+    // is made to 'intra_op_thread_count', which by default will take all
+    // threads
+    int intra_op_thread_count = -1;
+    err = ParseParameterInt(
```
tanmayv25, line 479:
```suggestion
    err = ParseParameter(
```

src/libtorch_utils.cc
```diff
@@ -149,6 +149,19 @@ ParseParameter(
   return nullptr;
 }
 
+TRITONSERVER_Error*
+ParseParameterInt(
```
tanmayv25, line 153:
```suggestion
ParseParameter(
```

src/libtorch_utils.h
```diff
@@ -62,4 +62,11 @@ TRITONSERVER_Error* ParseParameter(
     triton::common::TritonJson::Value& params, const std::string& mkey,
     bool* value);
 
+// If the key 'mkey' is present in 'params' then update 'value' with the
+// value associated with that key. If 'mkey' is not present in 'params' then
+// 'value' is set to 'default_value'.
+TRITONSERVER_Error* ParseParameterInt(
```
tanmayv25, line 68:
```suggestion
TRITONSERVER_Error* ParseParameter(
```

src/libtorch.cc
```diff
@@ -465,6 +471,54 @@ ModelState::ParseParameters()
...
+             Name() + "'")
+                .c_str());
+      }
+    }
+
+    // If "INTER_OP_THREAD_COUNT" is not present in 'parameters' then no update
+    // is made to 'inter_op_thread_count', which by default will take all
+    // threads
+    int inter_op_thread_count = -1;
+    err = ParseParameterInt(
```
tanmayv25, line 503:
```suggestion
    err = ParseParameter(
```

src/libtorch.cc
tanmayv25, line 1: We can use function overloading on ParseParameter.
Additionally, can you update the copyright years? 2023 -> 2024?

src/libtorch_utils.cc
tanmayv25, line 1: Update the copyright years.

src/libtorch_utils.h
tanmayv25, line 1: Update the copyright years.

tanmayv25 (changes requested):

yongbinfeng:
> Triton CLA

I think I've done already that, through my affiliation (fermilab) and my affiliation email. (The other PR is already merged: https://github.com/triton-inference-server/pytorch_backend/pull/120 so hopefully it should be fine I guess?)

tanmayv25 (approved):

tanmayv25:
Thanks for your contribution!

PR #132, opened 2024-06-18
Add cusparseLt in the installation to support 24.06

tanmayv25:
Adding cusparseLt in the backend installation. Initiated talks to include cusparseLt in the base container if possible.


nv-kmcgill53 (approved):
Thanks for the fix!

PR #137, opened 2024-08-13
build: RHEL8 PyTorch Backend

fpetrini15:
**Goal:** Support the PyTorch Backend on RHEL8 systems. In sum, these changes collect artifacts from the base and wheel containers necessary to build the Triton backend. Note, that these changes will only support a vanilla build of the PyTorch backend, with full functionality to be supported later.

Server Changes: https://github.com/triton-inference-server/server/pull/7568
TensorRT Backend: https://github.com/triton-inference-server/tensorrt_backend/pull/98
TensorFlow: https://github.com/triton-inference-server/tensorflow_backend/pull/105
ORT Backend: https://github.com/triton-inference-server/onnxruntime_backend/pull/266


CMakeLists.txt
```diff
@@ -208,61 +222,125 @@ set(OPENCV_LIBS
...
+      #   include/torchvision
+      # COMMAND ${CMAKE_COMMAND} -E make_directory "include/torchvision"
+      COMMAND ${CMAKE_COMMAND} -E make_directory "include"
+      COMMAND docker pull ${TRITON_PYTORCH_DOCKER_IMAGE}
+      COMMAND docker rm pytorch_backend_ptlib || echo "error ignored..." || true
+      COMMAND docker create --name pytorch_backend_ptlib ${TRITON_PYTORCH_DOCKER_IMAGE}
+      # Transfer and install torch wheel which contains many of the target libraries
+      COMMAND docker cp pytorch_backend_ptlib:/opt/transfer/. /tmp/ && /bin/sh -c "pip install /tmp/torch-*.whl"
+      COMMAND /bin/sh -c "for i in ${LIBTORCH_LIBS_STR} ; do echo copying $i && docker cp -L pytorch_backend_ptlib:${LIBTORCH_PATH}/$i $i ; done"
+      COMMAND cp ${LIBTORCH_PATH}/python3.10/site-packages/torch/lib/libc10.so libc10.so
```
mc-nv, line 243: can we use `Python3_SITEARCH` as it described in documentation? 
https://cmake.org/cmake/help/latest/module/FindPython3.html

CMakeLists.txt
```diff
@@ -208,61 +222,125 @@ set(OPENCV_LIBS
...
+      COMMAND cp ${LIBTORCH_PATH}/python3.10/site-packages/torch/lib/libc10_cuda.so libc10_cuda.so
+      COMMAND cp ${LIBTORCH_PATH}/python3.10/site-packages/torch/lib/libtorch.so libtorch.so
+      COMMAND cp ${LIBTORCH_PATH}/python3.10/site-packages/torch/lib/libtorch_cpu.so libtorch_cpu.so
+      COMMAND cp ${LIBTORCH_PATH}/python3.10/site-packages/torch/lib/libtorch_cuda.so libtorch_cuda.so
+      COMMAND cp ${LIBTORCH_PATH}/python3.10/site-packages/torch/lib/libtorch_cuda_linalg.so libtorch_cuda_linalg.so
+      COMMAND cp ${LIBTORCH_PATH}/python3.10/site-packages/torch/lib/libtorch_global_deps.so libtorch_global_deps.so
+      COMMAND cp ${LIBTORCH_PATH}/python3.10/site-packages/torch/lib/libcaffe2_nvrtc.so libcaffe2_nvrtc.so
```
mc-nv, line 250: Can we use `Python3_SITEARCH` as it described in documentation?
https://cmake.org/cmake/help/latest/module/FindPython3.html

mc-nv: if the list of the libs build depend on between Debian and RedHat  is equal. 
Ten only portion evaluation should resolve the path that will be used for copy, isn't?

fpetrini15: If I am understanding you correctly, you would like there to be no if/else clause since they share similar dependencies? This is something I wanted to avoid as well, however, the issue here is that for RHEL the files will be local (because it's where we install the torch wheel), whereas, for Debian, the image we use has the wheel pre-installed so we copy it out from the image.

That said, I default `${LIBTORCH_PATH}` to `/usr/local/lib`, so I have the opportunity to clean up some of the debian paths below by prefixing with `${LIBTORCH_PATH}`. I'll have to look into `Python3_SITEARCH` to see how it could factor in.

CMakeLists.txt
```diff
@@ -208,61 +220,119 @@ set(OPENCV_LIBS
...
+      COMMAND /bin/sh -c "if [ -f libmkl_avx512.so.1 ]; then patchelf --add-needed libmkl_core.so.1 libmkl_avx512.so.1; fi"
+      COMMAND /bin/sh -c "if [ -f libmkl_vml_def.so.1 ]; then patchelf --add-needed libmkl_gnu_thread.so.1 libmkl_vml_def.so.1; fi"
+      COMMAND /bin/sh -c "if [ -f libmkl_vml_def.so.1 ]; then patchelf --add-needed libmkl_intel_thread.so.1 libmkl_vml_def.so.1; fi"
+      COMMAND /bin/sh -c "if [ -f libmkl_vml_def.so.1 ]; then patchelf --add-needed libmkl_core.so.1 libmkl_vml_def.so.1; fi"
+      COMMAND /bin/sh -c "if [ -f libmkl_intel_thread.so.1 ]; then patchelf --add-needed libmkl_intel_lp64.so.1 libmkl_intel_thread.so.1; fi"
+      COMMAND docker rm pytorch_backend_ptlib
+      COMMENT "Extracting pytorch and torchvision libraries and includes from ${TRITON_PYTORCH_DOCKER_IMAGE}"
+      VERBATIM
+    )
+  endif() # RHEL_BUILD
```
mc-nv, line 335: May it worth to use expressions instead of `if|else` ?
https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

CMakeLists.txt
fpetrini15: Tried to leverage `FindPython` and oddly most of the results variables are blank:
```
Python3_STDARCH
Python3_STDLIB
Python3_SITELIB
Python3_SITEARCH
Python3_INCLUDE_DIRS /opt/_internal/cpython-3.10.13/include/python3.10 <-- only value retrieved
Python3_LIBRARIES
Python3_LIBRARY_DIRS
```

CMakeLists.txt
fpetrini15: I condensed things back down and removed the if/else. We now leverage conditional paths and some generator expressions.

CMakeLists.txt
```diff
@@ -84,7 +84,19 @@ else()
 endif()
 
 # Python.h needed by torch headers.
-find_package(Python3 REQUIRED COMPONENTS Development)
+find_package(Python3 REQUIRED COMPONENTS Development.Module)
+
+set(RHEL_BUILD OFF)
+set(LIB_DIR "lib")
+set(PY_INSTALL_PATH "/usr/local/lib/python3.10/dist-packages")
```
mc-nv, line 91: I'm not insist on it but with it we shouldn't carry about versions match as long as we aligned with DLFW.
```suggestion
set(PY_INSTALL_PATH ${Python_SITEARCH})
```

fpetrini15: Right, and I've checked that `sysconfig.get_path('platlib')` returns a value--and it does. However, CMake still fails to set the return values for some reason.

That said, after more consideration, I don't think we can go with this approach because the `PY_INSTALLPATH` we are setting is relative to the pytorch container we are copying files out of, not the container we are in when we are building. To illustrate why this is an issue, even if CMake was properly setting the value of `Python_SITEARCH`, relative to the build container, it would be:
`/opt/python/cp310-cp310/lib/python3.10/site-packages`
However, relative to the pytorch container it would need it to be:
`/opt/_internal/cpython-3.10.13/lib/python3.10/site-packages`

mc-nv (approved):
Commenting for the record. 
Change is OK. 
We may wan't to address some configuration in future, in context of how we building the PyTorch backend

CMakeLists.txt
```diff
@@ -220,35 +232,35 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
...
-    COMMAND docker cp pytorch_backend_ptlib:/usr/local/lib/python3.10/dist-packages/torch/lib/libtorch.so libtorch.so
-    COMMAND docker cp pytorch_backend_ptlib:/usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cpu.so libtorch_cpu.so
-    COMMAND docker cp pytorch_backend_ptlib:/usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cuda.so libtorch_cuda.so
-    COMMAND docker cp pytorch_backend_ptlib:/usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_cuda_linalg.so libtorch_cuda_linalg.so
-    COMMAND docker cp pytorch_backend_ptlib:/usr/local/lib/python3.10/dist-packages/torch/lib/libtorch_global_deps.so libtorch_global_deps.so
-    COMMAND docker cp pytorch_backend_ptlib:/usr/local/lib/python3.10/dist-packages/torch/lib/libcaffe2_nvrtc.so libcaffe2_nvrtc.so
+    COMMAND /bin/sh -c "for i in ${LIBTORCH_LIBS_STR} ; do echo copying $i && docker cp -L pytorch_backend_ptlib:$<IF:$<BOOL:${RHEL_BUILD}>,/opt/_internal/cpython-3.10.13/lib/$i,/usr/local/lib/$i> $i ; done"
+    COMMAND docker cp pytorch_backend_ptlib:${PY_INSTALL_PATH}/torch/lib/libc10.so libc10.so
```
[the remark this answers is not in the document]
fpetrini15: Taking a look at the docs, I don't believe so. These changes only impact the docker build flow--so we can rule out any functional changes to the local build. Additionally, "under-the-hood" these changes should leave the ubuntu flow exactly how it was prior to my edits. The only difference is these hardcoded paths will be centralized into variables. I believe CI gives us confidence regarding these claims. Is there a spot in particular you had concern about?

nv-kmcgill53 (approved):

PR #142, opened 2024-11-15
test: Upgrade Miniconda version for ubuntu 24.04

krishung5:
Fix L0_pytorch_python_runtime--PyTorch test.


krishung5:
The pytorch version installed by miniconda will be 2.5.1, which has torch compile support. It should resolve the previous issue: https://github.com/pytorch/pytorch/issues/120233#issuecomment-2089257409.
CC @kthui

mc-nv (approved):

PR #148, opened 2025-02-14
Fix RHEL build

pvijayakrish:


CMakeLists.txt
```diff
@@ -215,9 +230,21 @@ set(OPENCV_LIBS
...
+    set(OPENCV_LIBS
+        ${OPENCV_LIBS}
+        "libjpeg.so.62"  # RHEL uses only .so.62 version
+        "libpng16.so.16" # RHEL uses .so.16 version
+    )
+else()
+    set(OPENCV_LIBS
+        ${OPENCV_LIBS}
+        "libjpeg.so"     # Non-RHEL uses both .so and .so.62
+        "libjpeg.so.62"
```
mc-nv, line 244: Why two libraries? I'm afraid it can be same file (resolved by symbolic links) isn't?
If it's different files then question arise because these two lib may have same methods within and proper precedence may require.

pvijayakrish: Thanks @mc-nv for catching this. For non RHEL build it should only be libjpeg.so.

CMakeLists.txt
```diff
@@ -78,8 +78,16 @@ else()
...
+  if(TRITON_PYTORCH_ENABLE_TORCHVISION)
+    if(${RHEL_BUILD})
+      if(NOT EXISTS "${TRITON_PYTORCH_LIB_PATHS}/libtorchvision.so")
+        message(WARNING "TRITON_PYTORCH_ENABLE_TORCHVISION is on, but TRITON_PYTORCH_LIB_PATHS does not contain Torchvision package")
+      endif()
+    else()
+      if(NOT EXISTS "${TRITON_PYTORCH_LIB_PATHS}/libtorchvision.so.1")
+        message(WARNING "TRITON_PYTORCH_ENABLE_TORCHVISION is on, but TRITON_PYTORCH_LIB_PATHS does not contain Torchvision package")
+      endif()
+    endif()
```
mc-nv, line 90: Original statement wasn't written right. Expanding that statement in such a way may be not the best way. 
Instead it should be:
```suggestion
  find_library( LIBTORCHVISION libtorchvision.so libtorchvision.so.1 PATHS ${TRITON_PYTORCH_LIB_PATHS} )
  if(NOT ${LIBTORCHVISION}) 
    message(WARNING "TRITON_PYTORCH_ENABLE_TORCHVISION is on, but TRITON_PYTORCH_LIB_PATHS does not contain Torchvision package")
  endif(NOT ${LIBTORCHVISION})
```

CMakeLists.txt
```diff
@@ -162,10 +170,17 @@ set(PT_LIBS
...
+    set(PT_LIBS
+        ${PT_LIBS}
+        "libtorchvision.so"
+    )
+  else()
+    set(PT_LIBS
+        ${PT_LIBS}
+        "libtorchvision.so.1"
+    )
+  endif()
```
mc-nv, line 183: Using [generator-expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html) is more preferable in my opinion: 
```suggestion
  set(PT_LIBS
         ${PT_LIBS}
         $<IF:$<BOOL:RHEL_BUILD>,libtorchvision.so,libtorchvision.so.1>
  )
```
see: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#conditional-expressions

CMakeLists.txt
```diff
@@ -215,9 +230,20 @@ set(OPENCV_LIBS
...
+        "libjpeg.so.62"  # RHEL uses only .so.62 version
+        "libpng16.so.16" # RHEL uses .so.16 version
+    )
+else()
+    set(OPENCV_LIBS
+        ${OPENCV_LIBS}
+        "libjpeg.so"
+        "libpng16.so"
+    )
+endif()
```
mc-nv, line 246: Using [generator-expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html) is more preferable in my opinion: 
```suggestion
  set(OPENCV_LIBS
         ${OPENCV_LIBS}
         $<IF:$<BOOL:RHEL_BUILD>,libjpeg.so.62,libjpeg.so>
         $<IF:$<BOOL:RHEL_BUILD>,libpng16.so.16,libpng16.so>
  )
```
see: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#conditional-expressions

CMakeLists.txt
```diff
@@ -382,9 +406,15 @@ if (${TRITON_PYTORCH_DOCKER_BUILD})
...
-        "${CMAKE_CURRENT_BINARY_DIR}/libtorchvision.so.1")
+    if(${RHEL_BUILD})
+      set(TRITON_PYTORCH_LIBS
+          ${TRITON_PYTORCH_LIBS}
+          "${CMAKE_CURRENT_BINARY_DIR}/libtorchvision.so")
+    else()
+      set(TRITON_PYTORCH_LIBS
+          ${TRITON_PYTORCH_LIBS}
+          "${CMAKE_CURRENT_BINARY_DIR}/libtorchvision.so.1")
+    endif()
```
mc-nv, line 417: Using [generator-expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html) is more preferable in my opinion: 
```suggestion
  set(TRITON_PYTORCH_LIBS
          ${TRITON_PYTORCH_LIBS}
     $<IF:$<BOOL:RHEL_BUILD>,${CMAKE_CURRENT_BINARY_DIR}/libtorchvision.so,${CMAKE_CURRENT_BINARY_DIR}/libtorchvision.so.1>)
  )
```
see: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#conditional-expressionsI

CMakeLists.txt
```diff
@@ -78,16 +78,10 @@ else()
...
-    else()
-      if(NOT EXISTS "${TRITON_PYTORCH_LIB_PATHS}/libtorchvision.so.1")
-        message(WARNING "TRITON_PYTORCH_ENABLE_TORCHVISION is on, but TRITON_PYTORCH_LIB_PATHS does not contain Torchvision package")
-      endif()
-    endif()
+  find_library( LIBTORCHVISION libtorchvision.so libtorchvision.so.1 PATHS ${TRITON_PYTORCH_LIB_PATHS} )
+  if(NOT ${LIBTORCHVISION}) 
+    message(WARNING "TRITON_PYTORCH_ENABLE_TORCHVISION is on, but TRITON_PYTORCH_LIB_PATHS does not contain Torchvision package")
+  endif(NOT ${LIBTORCHVISION})
   endif()
```
mc-nv, line 85: Is this `endif()`  needs to be here?

pvijayakrish: Removed it.

CMakeLists.txt
pvijayakrish: From my perspective, using if/else is more readable and evaluates during the configuration phase. However, I just noticed that we're adopting [generator expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html) across other backends outside of PyTorch. If we want to maintain uniformity in our approach, I can make the changes.

mc-nv: Feel free to ignore my suggestions and resolve conversations as needed.

mc-nv (approved):

mc-nv (approved):

PR #154, opened 2025-08-09
feat: add bfloat16

phuhung273:
Relates https://github.com/triton-inference-server/server/issues/7853


MatthieuToulemont:
This PR is a lifesaver, especially given that ONNX does not support BF16 in Triton. Torchscript and ONNX are kinda the only backend that can run complex if / else logic

PR #158, opened 2025-09-27
fix crash during torch model reload with inter op thread parameter set

snadampal:
torch allows setting the interop thread count only once and before any inter-op parallel work is started because it creates threadpool statically.
To follow this constraint, check the interop thread count and set only if it's not yet set.

### DIFF
--- a/src/libtorch.cc
+++ b/src/libtorch.cc
@@ -532,15 +532,17 @@ ModelState::ParseParameters()
         TRITONSERVER_ErrorDelete(err);
       }
     } else {
-      if (inter_op_thread_count > 0) {
+      if ((inter_op_thread_count > 0) &&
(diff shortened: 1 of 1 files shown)


src/libtorch.cc
```diff
@@ -532,15 +532,17 @@ ModelState::ParseParameters()
         TRITONSERVER_ErrorDelete(err);
       }
     } else {
-      if (inter_op_thread_count > 0) {
+      if ((inter_op_thread_count > 0) &&
+          (inter_op_thread_count != at::get_num_interop_threads()) &&
```
kpedro88, line 536: This can still allow a repeated call if two model config files set the number of interop threads to different values. (See #159 for a safer solution.)

snadampal: Good. thanks for fixing the multiple config file scenario as well!

PR #196, opened 2026-06-02
feat: Enable PyTorch2 Batching

mattwittwer:
#### What does the PR do?

Enables default (first‑dimension) batching and sequence batching for the AOT Inductor (PT2) backend. Fixes the output responder to scatter batched results per‑request (first_dim_batching), implements control‑input + implicit‑state validation/registration and adds an io_data alias mechanism so control/state tensors resolve via ordinal (INPUT__N), forward (ARGS[...]), or descriptive (<name>__<index>) names - matching pt1. 

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [ ] Related issues are referenced.
- [ ] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [x] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [ ] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ ] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

https://github.com/triton-inference-server/server/pull/8814

#### Test plan:

- CI Pipeline ID:

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
- closes GitHub issue: #xxx

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,15 +317,15 @@ output: [
 
 > [!NOTE]
 > Sequence batching with implicit state is supported for AOT Inductor compiled models.
-> The model's `forward` function must accept the sequence control tensors (e.g. `CONTROL_SEQUENCE_START`, `CONTROL_SEQUENCE_READY`) and any implicit state input as ordinary tensor arguments, and return any new state value as one of its outputs.
+> The model's `forward` function must accept the sequence control tensors (e.g. `CONTROL_SEQUENCE_START`, `CONTROL_SEQUENCE_READY`, `CONTROL_SEQUENCE_CORRID`) and any implicit state input as ordinary tensor arguments, and return any new state value as one of its outputs.
+> `CONTROL_SEQUENCE_CORRID` is supported when configured with a numeric `data_type` (e.g. `TYPE_INT32`, `TYPE_INT64`); a `TYPE_STRING` correlation id is not supported because the AOT Inductor runtime is tensor-only.
 > The control and state tensors may be addressed in the `config.pbtxt` using either the ordinal `INPUT__<index>` / `OUTPUT__<index>` naming convention or the descriptive `<name>__<index>` convention (e.g. `START__2`, `INPUT_STATE__1`, `OUTPUT_STATE__1`) used by the legacy LibTorch backend; the descriptive name's trailing index selects the corresponding ordinal input/output.
 > As with default batching, the PT2 archive must be exported with a dynamic first (batch) dimension on every input.
 
 > [!WARNING]
 > The following features are **not** yet supported for AOT Inductor compiled models packaged as a PT2 model archive:
-> * `CONTROL_SEQUENCE_CORRID` (typed sequence control); only the boolean controls `CONTROL_SEQUENCE_START`, `CONTROL_SEQUENCE_END`, and `CONTROL_SEQUENCE_READY` are supported.
+> * `TYPE_STRING` (`torch.export` byte/string) inputs and outputs, including string sequence state and a `TYPE_STRING` `CONTROL_SEQUENCE_CORRID`. The AOT Inductor runtime is tensor-only and cannot carry string values.
 > * Batching of models with complex (nested `dict`/`tuple`/`list`) inputs or outputs; default batching currently targets models whose inputs and outputs are plain tensors.
-> * Batching of `TYPE_STRING` (`torch.export` byte/string) inputs, including string sequence state.
 
 ### PyTorch 2.0 Models
 
--- a/README.md
+++ b/README.md
@@ -326,6 +326,7 @@ output: [
 > The following features are **not** yet supported for AOT Inductor compiled models packaged as a PT2 model archive:
 > * `TYPE_STRING` (`torch.export` byte/string) inputs and outputs, including string sequence state and a `TYPE_STRING` `CONTROL_SEQUENCE_CORRID`. The AOT Inductor runtime is tensor-only and cannot carry string values.
 > * Batching of models with complex (nested `dict`/`tuple`/`list`) inputs or outputs; default batching currently targets models whose inputs and outputs are plain tensors.
+> * RHEL (manylinux) builds. AOT Inductor is not currently supported on RHEL.
 
 ### PyTorch 2.0 Models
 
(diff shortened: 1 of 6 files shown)


src/pt2/io_data.hh
```diff
@@ -119,6 +119,16 @@ class io_data {
   ///                     For example, "INPUT__0" or "OUTPUT__0".
   void emplace(const std::string& name, const std::string& ordinal_name);
 
+  /// @brief Registers an additional name as an alias of an existing data
+  ///        specification. This allows the same underlying input/output to be
+  ///        addressed by more than one name, e.g. mapping the descriptive
+  ///        sequence-batching name "START__1" onto the ordinal name "INPUT__1".
```
whoisj: interesting, this goes against the prescriptive nature of our existing AOTI implementation, but I am open to it.

in what case would an input/output be referenced by more than a single name?

README.md
```diff
@@ -301,9 +301,31 @@ output: [
...
+>     model, sample_inputs, dynamic_shapes=({0: batch}, {0: batch})
+> )
+> torch._inductor.aoti_compile_and_package(exported_model, package_path="model.pt2")
+> ```
+
+> [!NOTE]
+> Sequence batching with implicit state is supported for AOT Inductor compiled models.
+> The model's `forward` function must accept the sequence control tensors (e.g. `CONTROL_SEQUENCE_START`, `CONTROL_SEQUENCE_READY`, `CONTROL_SEQUENCE_CORRID`) and any implicit state input as ordinary tensor arguments, and return any new state value as one of its outputs.
+> `CONTROL_SEQUENCE_CORRID` is supported when configured with a numeric `data_type` (e.g. `TYPE_INT32`, `TYPE_INT64`); a `TYPE_STRING` correlation id is not supported because the AOT Inductor runtime is tensor-only.
+> The control and state tensors may be addressed in the `config.pbtxt` using either the ordinal `INPUT__<index>` / `OUTPUT__<index>` naming convention or the descriptive `<name>__<index>` convention (e.g. `START__2`, `INPUT_STATE__1`, `OUTPUT_STATE__1`) used by the legacy LibTorch backend; the descriptive name's trailing index selects the corresponding ordinal input/output.
```
whoisj: 🤔 this gives us an interesting conundrum. the call needs to know the index of the input/output, but also needs to name it accordingly.

perhaps there's a secondary config.pbtxt field we could use to say something like `sequence_start: "INPUT__0"` or similar?

src/pt2/io_data.hh
mattwittwer: The only case was sequence control/state tensors — the motivation was LibTorch parity, so an existing LibTorch sequence config (which names controls/state descriptively, e.g. START__2 / INPUT_STATE__1) could be reused on AOTI by aliasing those onto the ordinal INPUT__<index> entry. 

Agreed it cuts against AOTI's prescriptive design, so I've dropped the alias and require ordinal INPUT__<index> / OUTPUT__<index> naming; the control's role now comes from the control_input kind.

README.md
mattwittwer: Resolved by going ordinal-only — control_input { name: "INPUT__2", control: [{ kind: CONTROL_SEQUENCE_START }] } already separates identity (name) from role (kind), so no new config field is needed. README updated.

whoisj: what happens if they're using the `ARGS[0]` style interface?

mattwittwer: The forward ARGS[...] / RESULT[...] interface works the same as the ordinal one — every input/output is registered under both its forward name and its INPUT__<index> / OUTPUT__<index> ordinal, so a control/state tensor resolves either way (e.g. name: "ARGS[2]" for START, state ARGS[1]→RESULT[1]). The control's role still comes from the control kind, not the name. Ref: https://github.com/triton-inference-server/pytorch_backend/blob/e37acc501604989be855b908bb31436dbebd5333/src/pt2/model_instance_state.cc#L112-L122

I added a test for this here: [test_forward_interface_sequence](https://github.com/triton-inference-server/server/blob/a0dc3d3738333481c92944675132a6b292a78e2b/qa/L0_torch_aoti/torch_aoti_infer_test.py#L498-L518). This runs the sequence model with all control/state addressed via the forward interface and verifies it matches the ordinal-named model's results.

src/pt2/model_instance_state.cc
```diff
@@ -1681,41 +1676,49 @@ ModelInstanceState::ValidateBooleanSequenceControl(
...
+{
+  // The control/state tensor must resolve to one of the model's inputs using
+  // the ordinal "INPUT__<index>" name or the forward-argument name.
+  if (!map_inputs_.contains(tensor_name)) {
+    THROW_TRITON_EXCEPTION(
+        TRITONSERVER_ERROR_INVALID_ARG,
+        "Sequence " << context << " \"" << tensor_name << "\" for model \""
+                    << Name()
+                    << "\" does not correspond to any model input. Sequence "
+                       "control and "
```
yinggeh: Formatting looks odd

yinggeh (dismissed):

pskiran1 (dismissed):

whoisj (dismissed):

src/pt2/model_instance_state.cc
mattwittwer: Combined the string on lines 1702-1704 for better formatting. 

Pre-commit/clang still prefers to leave 'Name()' on its own line to avoid another split in the string.

yinggeh (dismissed):

whoisj (dismissed):
LGTM

yinggeh (approved):

whoisj (approved):

PR #194, opened 2026-05-27
post: Update default branch post-26.05

mc-nv:
<sup>CI (internal): [#52829374](http://tritonserver.local/ci/pipelines/52829374)</sup>

#### What does the PR do?

Cherry-pick of `r26.05`#191 ("fix: streamline library distribution for 26.05 upstream container build") to `main`: removes an absent reference and changes library distribution out of the PyTorch image. Required so `main` can build against the `26.05` upstream container.

Changed files: `CMakeLists.txt`.

#### Checklist

- [x] I have made sure this PR is not a duplicate.
- [x] I have made sure the commit message is clear and follows [Conventional Commits](https://www.conventionalcommits.org/).
- [ ] I have added unit tests for any new functionality.
- [x] I have made sure all tests are passing.
- [ ] For documentation changes: I have previewed the changes and they render correctly.

#### Commit Type:

- [ ] `feat` (A new feature)
- [x] `fix` (A bug fix)
- [ ] `docs` (Documentation-only changes)
- [ ] `style` (Changes that do not affect the meaning of the code — formatting, whitespace, etc.)
- [ ] `refactor` (A code change that neither fixes a bug nor adds a feature)
- [ ] `perf` (A code change that improves performance)
- [ ] `test` (Adding missing tests or correcting existing tests)
- [x] `build` (Changes that affect the build system or external dependencies)
- [ ] `ci` (Changes to CI configuration files and scripts)
- [ ] `chore` (Other changes that don't modify src or test files)

#### Related PRs:

- triton-inference-server/server#8804
- triton-inference-server/model_analyzer#1024
- triton-inference-server/tensorrtllm_backend#857

Original (r26.05): triton-inference-server/pytorch_backend#191

#### Where should the reviewer start?

`CMakeLists.txt`

#### Test plan:

Cherry-pick of an already-merged fix on `r26.05`; verified against the 26.05 upstream container build.

- CI Pipeline ID: 52829374

#### Caveats:

Cherry-pick from `r26.05`.

#### Background

Part of TRI-1063 post-26.04 release maintenance: forward-port the 26.05 library-distribution fix so `main` builds cleanly against the 26.05 upstream container.

#### Related Issues:

Resolves: TRI-1063

PR #197, opened 2026-06-10
NV embedding layer loading

z52527:
## Problem

Some PT2/AOTI models need a one-time **native initialization** at model load — before the package is loaded — that cannot live inside `model.pt2`. A concrete example: a model whose custom ops resolve embedding tables from a process-global registry at execute time, where the weights are stored in sidecar files next to the package (not inside `model.pt2`). Loading the `.pt2` does not load them, so the **first inference request** fails:

```
... no binding registered for layer_id=1
-> Inductor model instance execution failure ... model_container_runner.cpp:150
```

The Triton PyTorch backend currently has no entry point to run such load-time initialization, so today it can only be done with fragile workarounds (e.g. an `LD_PRELOAD` shim that hijacks an unrelated CUDA call to piggyback on its timing).

## What this PR does

Adds a generic, optional **model-init hook** to the PT2 path. When a model's `config.pbtxt` sets the `MODEL_INIT_LIBRARY` parameter, the backend `dlopen()`s that shared library in `LoadModel()` — right before the `AOTIModelPackageLoader` is created (CUDA is already initialized there) — and calls a fixed C entry point. Design choices:

* **Backend stays generic** — it does not link against the library and knows nothing about what it does. It only `dlopen`s the `.so`, calls `triton_pytorch_model_init(model_dir, device_index)`, and releases it (`triton_pytorch_model_release` + `dlclose`) on unload. All model-specific logic lives in the user-provided plug-in, built and shipped downstream.
* **No new dependency** — the only added link is `${CMAKE_DL_LIBS}` (CMake's built-in libdl variable, for `dlopen`/`dlsym`/`dlclose`). The backend links no vendor libraries, so the stock `nvcr.io/nvidia/pytorch:26.05-py3` build is unaffected.
* **No-op by default** — a model without the parameter is completely unaffected; a `dlopen`/`dlsym` failure fails model load (fail-closed).
* **Scoped to the model lifetime** — the handle is held in `ModelState` and freed in the destructor.
* **Fixes a pre-existing parameters bug** — `ParseParameters()` was guarded by `if (!ModelConfig().Find("parameters", ...))`, so PT2 parameters were parsed only when the section was *absent*; any model that set a `parameters` block had them all (`INFERENCE_MODE`, …) silently ignored. Dropping the `!` makes them take effect.

The library must export:

```c
// Called once at model load, before the package is loaded.
void* triton_pytorch_model_init(const char* model_dir, int device_index);
// Optional; called once on unload with the handle returned above.
void  triton_pytorch_model_release(void* state);
```

Verified end-to-end on the official `nvcr.io/nvidia/tritonserver:26.05-py3` image with an AOTI model (parameter parsed → hook loaded → model READY → inference correct); also compiles against `main`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -503,6 +503,7 @@ target_link_libraries(
     triton-core-backendapi # from repo-core
     triton-core-serverstub # from repo-core
     triton-backend-utils   # from repo-backend
+    ${CMAKE_DL_LIBS}       # dlopen/dlsym/dlclose for the MODEL_INIT_LIBRARY hook
     ${TRITON_PYTORCH_LDFLAGS}
     ${TRITON_PYTORCH_LIBS}
     $<$<BOOL:${TRITON_PYTORCH_OPTREE}>:optree>
--- a/README.md
+++ b/README.md
@@ -305,6 +305,56 @@ output: [
 > Support for batch sizes greater than 1 and for sequence batching for AOT Inductor compiled models has not be completed.
 > These Triton Server features are currently unavailable for PyTorch models compiled using AOT Inductor and packaged as a PT2 model archive.
 
+### Model Initialization Hook (`MODEL_INIT_LIBRARY`)
+
+Some models need to run one-time **native** initialization at load time — before
+the model package is loaded — that cannot be expressed inside `model.pt2`. A
+typical case is a model whose custom operators resolve large weights from a
+process-global registry at execute time, where those weights live outside the
+package (e.g. in sidecar files next to `model.pt2`) and must be read into memory
+and registered first.
+
+The PT2 path supports an optional, per-model hook for this. When a model's
+`config.pbtxt` sets the `MODEL_INIT_LIBRARY` parameter, the backend `dlopen()`s
+that shared library at model load and calls its initialization entry point. The
+backend does **not** link against the library and knows nothing about what it
+does; it only loads it, calls a fixed entry point, and releases it on unload.
+
+```
+parameters {
+  key: "MODEL_INIT_LIBRARY"
+  value: { string_value: "/abs/path/to/libmy_model_init.so" }
+}
+```
+
+The library must export the following C entry point, and may optionally export a
+matching finalizer:
+
+```c
+// Called once at model load, before the model package is loaded.
+//   model_dir    : the versioned model directory (the one containing model.pt2)
+//   device_index : the GPU ordinal for the instance, or -1 for CPU
+// Returns an opaque handle that is passed back to the finalizer on unload, or
+// NULL if there is nothing to keep.
+extern "C" void* triton_pytorch_model_init(const char* model_dir, int device_index);
+
+// Optional. Called once on model unload with the handle returned above.
+extern "C" void  triton_pytorch_model_fini(void* state);
+```
+
+If `MODEL_INIT_LIBRARY` is unset (the default), no library is loaded and this is
+a complete no-op. If it is set but the library cannot be `dlopen()`ed or does not
+export `triton_pytorch_model_init`, model load fails with an error.
+
+> [!WARNING]
+> `MODEL_INIT_LIBRARY` makes the backend load and execute native code from the
+> path given in the model configuration, in the server process and with its
+> privileges. Only point it at libraries you trust. This is the same trust level
+> already required for the model repository itself — a PT2 package contains
+> compiled code that runs when the model is loaded — so it does not widen the
+> trust boundary, but the model repository must remain a trusted,
+> operator-controlled source.
+
 ### PyTorch 2.0 Models
 
 PyTorch 2.0 features are available.
--- a/src/pt2/model_state.cc
+++ b/src/pt2/model_state.cc
@@ -353,9 +353,9 @@ ModelState::LoadModel(
 
   // Optional per-model native init hook (MODEL_INIT_LIBRARY parameter). Lets a
   // model run one-time native setup before its package is loaded -- e.g.
-  // loading external embedding weights into a process-global registry -- without
-  // the backend linking against or knowing anything about that library. No-op
-  // when the parameter is unset.
+  // loading external embedding weights into a process-global registry --
+  // without the backend linking against or knowing anything about that library.
+  // No-op when the parameter is unset.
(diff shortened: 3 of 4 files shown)


CMakeLists.txt
mc-nv, line 339: Same change was introduced in https://github.com/triton-inference-server/pytorch_backend/pull/194 and may require rebase.

CMakeLists.txt
```diff
@@ -630,3 +630,49 @@ export(
...
+# ===========================================================================
+if(${TRITON_ENABLE_GPU})
+  enable_language(CUDA)
+  set(NVE_ROOT "/workspace/deps/nve" CACHE PATH "nv-embedding-cache source root")
+  set(NVE_LIB_DIR "/workspace/nve_libs" CACHE PATH "dir containing libnve-*.so")
+
+  find_library(NVE_TORCH_OPS_LIB nve-torch-ops
+    PATHS ${NVE_LIB_DIR} NO_DEFAULT_PATH REQUIRED)
+  find_library(NVE_COMMON_LIB nve-common
+    PATHS ${NVE_LIB_DIR} NO_DEFAULT_PATH REQUIRED)
```
mc-nv: Is it possible to get more clarity on this libraries and where they will be distributed from. 
I can't find relevant library inside `nvcr.io/nvidia/pytorch:26.05-py3` which is source of dependency for "Triton PyTorch Backend", which will cause build failure in case of accepting this change.

README.md
whoisj, line 126: Looks like your changes need to be rebased onto latest `main` before they can be merged.

whoisj (changes requested):

src/pt2/model_state.cc
```diff
@@ -696,6 +705,29 @@ ModelState::ParseParameters()
       }
       TRITONSERVER_ErrorDelete(err);
     }
+    {
+      TritonJsonValue init_lib_param;
+      if (parameters.Find("MODEL_INIT_LIBRARY", &init_lib_param)) {
```
whoisj, line 710: The one change I'd make here would be to ensure that the lib file was located within the model repository (that's our standard rule for loading files from local disk), but I can live with this as-is for now. I'll take an action item to add the check later.

@z52527, would requiring the file to be inside Triton's model repository be a breaking change for you?

src/pt2/model_state.cc
```diff
@@ -720,6 +752,68 @@ ModelState::ParseParameters()
...
+                               << Name() << "\" (device_index=" << device_index
+                               << ").");
+}
+
+ModelState::~ModelState()
+{
+  if (model_init_dl_handle_ != nullptr) {
+    using ModelFiniFn = void (*)(void*);
+    auto* fini_fn = reinterpret_cast<ModelFiniFn>(
+        dlsym(model_init_dl_handle_, "triton_pytorch_model_fini"));
```
whoisj, line 809: "triton_pytorch_model_fini" is an odd name. Perhaps "triton_pytorch_model_release" or "triton_pytorch_model_close" would make more sense?

whoisj (dismissed):
Overall, this is good. Thank you for the contribution. I've left a couple of comments but am willing to accept this change as-is.

src/pt2/model_state.cc
z52527: Done. Renamed to "triton_pytorch_model_release". Thanks!

src/pt2/model_state.cc
z52527: I think it's fine as a follow-up. We'd just keep the .so inside a model directory rather than at the repo root.

CMakeLists.txt
z52527: Resolved in the latest revision, the backend no longer links NVE; it dlopen()s a user-provided plug-in via the new MODEL_INIT_LIBRARY hook, so nothing NVE-specific ships in the backend image.

CMakeLists.txt
```diff
@@ -503,6 +503,7 @@ target_link_libraries(
     triton-core-backendapi # from repo-core
     triton-core-serverstub # from repo-core
     triton-backend-utils   # from repo-backend
+    ${CMAKE_DL_LIBS}       # dlopen/dlsym/dlclose for the MODEL_INIT_LIBRARY hook
```
mc-nv, line 506: `CMAKE_` prefix is a reserved keyword, afraid it's usage may create some confusion. See examples below.
http://cmake.org/cmake/help/latest/manual/cmake-variables.7.html

Also change description explains following, which means those libraries are vital to preserve the functionality, isn't?
> An HSTU AOTI package (model.pt2) calls nve_ops::embedding_lookup(keys, layer_id),
which looks the embedding table up by layer_id in a process-global
NVELayerRegistry. The embedding weights do not live inside model.pt2 —
they are separate files next to it (<model_dir>/metadata.json +
<model_dir>/weights/*.nve). So loading the .pt2 does not load them, and the
first inference request fails

Could you please share the missed libraries origin ?

whoisj:
CI Pipeline ID: 55028851

CMakeLists.txt
whoisj: @z52527 please follow up with @mc-nv's requests. This is blocking merge.

whoisj:
I've pushed a change to make the pre-commit hook happy. @z52527, please pull your branch to receive my contribution before working on it to avoid conflicts; or duplicate my work in your own commit. Thanks.

CMakeLists.txt
z52527: Sorry for the confusion. PR description was stale; it still described the earlier "link NVE into the backend" version. I've updated it.
In the current revision the backend links no NVE or vendor libraries. There are no missing libraries to source. The only CMake change is target_link_libraries(... ${CMAKE_DL_LIBS}). CMAKE_DL_LIBS is a built-in CMake variable for the platform's dynamic-loading library (libdl), needed for the dlopen/dlsym/dlclose the hook uses.

whoisj:
@z52527 this PR is currently blocked by CI failures unrelated to this PR, that make it nearly impossible for me to validate this PR. I'll return to this PR as soon as those test failures have been resolved. Apologies for the delay.

whoisj:
CI Pipeline ID: 55960871

whoisj (approved):

PR #205, opened 2026-08-13
fix: add guard to aoti package loader

mattwittwer:
#### What does the PR do?

Serializes `torch::inductor::AOTIModelPackageLoader` construction in the pt2 path behind a process-wide mutex.

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [x] Related issues are referenced.
- [ ] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [ ] Added [test plan](#test-plan) and verified test passes.
- [x] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [x] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

62435323

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
- closes GitHub issue: #xxx

### DIFF
--- a/src/pt2/model_state.cc
+++ b/src/pt2/model_state.cc
@@ -47,6 +47,12 @@
 namespace {
 std::once_flag pytorch_interop_threads_flag;
 std::once_flag pytorch_intraop_threads_flag;
+
+// AOTIModelPackageLoader construction is not thread-safe: libtorch's
+// load_json_file() (model_package_loader.cpp) parses into a shared static
+// nlohmann::json, so concurrent package loads corrupt each other's metadata.
+// Serialize construction until this is fixed upstream.
+std::mutex aoti_package_loader_ctor_mutex{};
 }  // namespace
 
 namespace triton::backend::pytorch::pt2 {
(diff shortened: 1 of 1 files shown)


whoisj:
Can only a single model be loaded safely at a time?

mattwittwer:
> Can only a single model be loaded safely at a time?

@whoisj - Good clarification, model loading overall is still parallel. The mutex covers only the construction of torch::inductor::AOTIModelPackageLoader. Non-AOTI models are unaffected, and for AOTI models everything after construction (call-spec parsing, instance setup) runs outside the lock.

The issue is in libtorch here: [torch/csrc/inductor/aoti_package/model_package_loader.cpp](https://github.com/pytorch/pytorch/blob/6a1e400f48/torch/csrc/inductor/aoti_package/model_package_loader.cpp#L245) this parses each package's metadata into a function-local static nlohmann::json and returns a reference to it, so two loaders constructing concurrently corrupt each other's view.

whoisj (approved):

